PR #2, opened 2015-01-22
producer::config takes service_name for notify

fessyfoo:
Note this was not tested.  posting as explaination of the issue.  @liamjbennett  let me know what you think

if kafka::mirror is used and  service_restart  is not set to false then a
catalog compilation error occurs.  because producer::config will setup a
notify to a non existant service.

File[/opt/kafka/config/producer.properties] { notify => Service[kafka]
}, because Service[kafka] doesn't seem to be in the catalog

so we allow producer::config to take a service_name parameter so that when
it is used by kafka::mirror::config the correct kafka-mirror service name
can be passed.

### DIFF
--- a/manifests/mirror/config.pp
+++ b/manifests/mirror/config.pp
@@ -19,7 +19,8 @@ class kafka::mirror::config(
 
   class { 'kafka::producer::config':
     config          => $producer_config,
-    service_restart => $service_restart
+    service_restart => $service_restart,
+    service_name    => 'kafka-mirror'
   }
 
   create_resources('kafka::consumer::config', $consumer_config)
--- a/manifests/producer/config.pp
+++ b/manifests/producer/config.pp
@@ -9,13 +9,14 @@
 #
 class kafka::producer::config(
   $config = {},
-  $service_restart = $kafka::producer::service_restart
+  $service_restart = $kafka::producer::service_restart,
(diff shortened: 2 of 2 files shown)


ffrank:
Are `mirror` and plain `kafka` mutually exclusive? If so, we should set up an alias so that the notify can go for that.

Otherwise, we should tag both services and notify using a collector.

    notify => Service<| tag == 'kafka-service-resource-tag' |>

Perhaps we should do this either way - it should cover all bases and I cannot see it causing harm.

daenney:
@liamjbennett @ffrank @fessyfoo Can you hash this out? The original PR is from almost a year ago but it still seems sensible to me. If we deem it no longer necessary then please close the PR :smile:.

jyaworski:
Hello:

What's the status of this PR?

ffrank:
Just stared at it again and a little confused.

The opening comment states that `Service[kafka]` is missing from the catalog in the `::mirror` case, but the patch will target `Service['kafka-producer']` in the default case. Seeing as this is apparently untested, I'm a little on the edge.

The idea is sound though, so I'll see about following this up with a fixup branch so that this can be merged anyhow.

dhoppe:
@ffrank Give me some minutes. I am just working on it, because there are some more things broken.

PR #7, opened 2015-03-30
Make beaker run on centos

bjoernhaeuser:
Also I cleaned up all things that I encountered. I synchronized the
beaker nodesets with the support platforms in the readme.


bjoernhaeuser:
CLA signed.

Issue #9, opened 2015-05-10
Both "version" and "scala_version" are ignored within "kafka::broker"


loki2302:
```puppet
class { 'kafka::broker':
  version => '0.8.2.1',
  scala_version => '2.10',
  install_java => true
}
```
Instead of 2.10 there's 2.8.0 and instead of 0.8.2.1 there's 0.8.1.1:
```
...
Notice: /Stage[main]/Kafka/File[/var/log/kafka]/ensure: created
Notice: /Stage[main]/Kafka/File[/opt/kafka-2.8.0-0.8.1.1]/ensure: created
Notice: /Stage[main]/Kafka/File[/opt/kafka]/ensure: created
...
```

dominics:
This issue is not caused by the version being *ignored*. The declaration given above correctly passes the versions to the `::kafka` init class. (If you put a `notify` in there, you'll see it received the correct values.)

What's actually happening is the `$install_dir`/`$install_directory` is incorrectly handled. If you don't specify it to `kafka::broker`, then it uses `$kafka::params::install_dir` (which is always a static default value of `/opt/kafka-2.8.0-0.8.1.1` - built in `params.pp`). This value is eventually passed to `::kafka`, which means the symlink isn't changed, and the downloaded package is never untar'd.

Basically [this line right here](https://github.com/puppet-community/puppet-kafka/blob/master/manifests/broker.pp#L51) makes no sense at all. 

(Yaaaaay, [another case](https://www.devco.net/archives/2013/12/09/the-problem-with-params-pp.php) of a "params class" causing confusion and breaking things.)

PR #10, opened 2015-05-12
Don't use params.pp value for install_dir

dominics:
See issue #9 for details of the issue.

To see why this is the right fix, consider someone using `kafka::broker` and specifying the versions of Kafka and Scala to use. At the moment, unless they specify the install directory as well, the default static value of `/opt/kafka-2.8.0-0.8.1.1` is used (i.e. `$kafka::params::install_dir`). This will be mismatched to whatever version they're trying to install.

(Practically, it's worse - it causes the symlink at `/opt/kafka` to be wrong, and it refuses to untar the package because the `creates` condition on the untar `exec` succeeds under the incorrect directory.)

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -48,7 +48,7 @@
 class kafka::broker (
   $version = $kafka::params::version,
   $scala_version = $kafka::params::scala_version,
-  $install_dir = $kafka::params::install_dir,
+  $install_dir = '',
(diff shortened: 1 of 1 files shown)


dhoppe:
@dominics: Thank you very much for your contribution. This issue has been resolved by commit ae994d5.

Kind regards,
Dennis

PR #11, opened 2015-05-19
Make the user broker service will run as configurable.

tmclaugh:
The puppet-kafka module uses sudo to launch all it's services but launches them all as root.  This doesn't make sense since the script can only be run as root as root typically is the only user with access to /var/run/ where the service's PID file is dropped.  Additionally, the module creates a kafka user and group.  Why bother creating the group when the service and data will be owned by root?

This is a first pass to keep my local work moving and to spark discussion about the correct way to fix this.

Note: Did not file a Jira for this because component does not exist.

### DIFF
--- a/templates/init.erb
+++ b/templates/init.erb
@@ -24,7 +24,7 @@ start() {
         start
       fi
     else
-        exec sudo KAFKA_JMX_OPTS="$KAFKA_JMX_OPTS" $DAEMON $DAEMON_OPTS >> /var/log/kafka/server.log 2>&1 &
+        exec sudo -u <%= service_user %> -g <%= service_group %> KAFKA_JMX_OPTS="$KAFKA_JMX_OPTS" $DAEMON $DAEMON_OPTS >> /var/log/kafka/server.log 2>&1 &
         sleep 2
         PID=`ps ax | grep -E '[k]afka.Kafka' | awk '{print $1}'`
         echo $PID > $PID_FILE;
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -37,6 +37,12 @@
 # [*service_restart*]
 # Boolean, if the configuration files should trigger a service restart
 #
+# [*service_user*]
+# Name of user to run service as
+#
+# [*service_group*]
+# Name of group to run service as
+#
 # === Examples
 #
 # Create a single broker instance which talks to a local zookeeper instance.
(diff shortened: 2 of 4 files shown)


jyaworski:
Hello:

What's the status of this PR? Can it be rebased?

tmclaugh:
Okay, let me try and get to this.  I have a few outstanding upstream PRs from this employer I should probably finish off in the coming 2 weeks.

danzilio:
slacker

jyaworski:
OK, thanks @tmclaugh

bastelfreak:
@tmclaugh how is the current state here?

templates/init.erb
```diff
@@ -24,7 +24,7 @@ start() {
         start
       fi
     else
-        exec sudo KAFKA_JMX_OPTS="$KAFKA_JMX_OPTS" $DAEMON $DAEMON_OPTS >> /var/log/kafka/server.log 2>&1 &
+        exec sudo -u <%= service_user %> -g <%= service_group %> KAFKA_JMX_OPTS="$KAFKA_JMX_OPTS" $DAEMON $DAEMON_OPTS >> /var/log/kafka/server.log 2>&1 &
         sleep 2
         PID=`ps ax | grep -E '[k]afka.Kafka' | awk '{print $1}'`
```
igalic: `pgrep -f kafka.Kafka`

igalic:
my main concern with this thing is: now that we're correcting the behaviour, and not running as root anymore, how will kafka be able to write its state?

bastelfreak:
HI @tmclaugh, how is the current state here, could you rebase?

IMO this is a breaking change. Could you also add a note to the changelog for this?

igalic:
hrm… i should incorporate this change…
it should also be "ported" over to systemd.

igalic:
hrm… our current `su` approach seems to work as intended:
```
root@kafka2-elk0:~# ps -p 4380 -o user,euser,ruser,group,egroup,rgroup,cmd
USER     EUSER    RUSER    GROUP    EGROUP   RGROUP   CMD
kafka    kafka    kafka    kafka    kafka    kafka    java -Xmx1G -Xms1G -server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeap
```

tmclaugh:
Sorry, this PR was filed from my last job.  Didn't have time to look back after this PR (startup fun) and now onto a new job as of a few months ago with no means of testing things from this anymore.

PR #14, opened 2015-06-25
Update topic.pp

VuokkoVuorinnen:
The unless statement matched too much topics for example, if you had a 'foo' and 'foobar' topic, deleted the 'foo' topic and tried to recreate it, puppet wouldnt do it because '*foo*' still matches 'foobar'

### DIFF
--- a/manifests/broker/topic.pp
+++ b/manifests/broker/topic.pp
@@ -17,7 +17,7 @@ define kafka::broker::topic(
   if $ensure == 'present' {
     exec { "create topic ${name}":
       command => "/opt/kafka/bin/kafka-topics.sh --create --zookeeper '${zookeeper}' --replication-factor ${replication_factor} --partitions ${partitions} --topic ${name}",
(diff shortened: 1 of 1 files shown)


nibalizer:
I believe only egrep can take the regex syntax. Can you confirm?

VuokkoVuorinnen:
Tested it on CentOS and now on my Ubuntu laptop:

````
patrick@sinnlig:~ $ echo "foobar" | grep '^foo$'
patrick@sinnlig:~ $ echo $?
1
patrick@sinnlig:~ $ echo "foo" | grep '^foo$'
foo
patrick@sinnlig:~ $ echo $?
0
````

Seems to also work with normal grep

nibalizer:
Ah yes, Sorry.

I don't really know a whole lot about kafka and zookeeper. I'll leave a +1 on this, if it doesn't get more activity in a few days I'll sweep around and merge it.

@VuokkoVuorinnen if you're using this module, would you be willing to write some simple beaker tests so that this kind of change can land with more confidence?

dhoppe:
@VuokkoVuorinnen @nibalizer: Thank you very much for your contribution. This issue has been resolved by commit ae994d5.

Kind regards,
Dennis

PR #15, opened 2015-06-26
Make wget installation optional

danieldreier:
Add an install_wget parameter to allow users to entirely disable management of wget. The require on an exec will still expect the resource to be exist, so it must be provided via another module if install_wget is set to false.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -67,7 +67,11 @@ class kafka (
     }
   }
 
-  ensure_resource('package','wget', { 'ensure' => 'present' })
+  if ! defined(Package['wget']) {
+    package {'wget':
+      ensure => present
+    }
+  }
 
   group { 'kafka':
     ensure => present


danieldreier:
To clarify my use case here, we have wget set to `ensure => latest`, so the `create_resource` wrapping wget in this module doesn't detect that as an identical resource and attempts to create a wget package resource, which conflicts with the one we declared previously.

nibalizer:
Are you sure this is what you wanted to push up? We've added a parameter that, as far as I can tell, doesn't do any thing. Unless ensure_resource knows how to look up the install_wget variable, which would be vomit inducing.

danieldreier:
@nibalizer you're right, I forgot to include the relevant conditional. I actually changed the approach and switched out the parameter to just check for resource conflicts using `if ! defined` instead of relying on ensure_resource.

igalic:
what does puppet-kafka (or kafka?) use wget for?

PR #22, opened 2015-10-13
Runuser doesn't exist in Ubuntu12.*/14.*, using su instead

victorgp:
Runuser comes with version 2.23 of util-linux which doesn't exist
in ubuntu12.* or 14.* repositories so a backport or custom compilation is needed

I don't think is worth breaking the Ubuntu compatibility of this
puppet module just because of this, so using "su" instead of runuser
is more appropiate

### DIFF
--- a/templates/init.erb
+++ b/templates/init.erb
@@ -32,7 +32,7 @@ start() {
       start
     fi
   else
-    /sbin/runuser $KAFKA_USER -c "KAFKA_JMX_OPTS=\"$KAFKA_JMX_OPTS\" $DAEMON $DAEMON_OPTS > /var/log/kafka/server.out 2> /var/log/kafka/server.err &"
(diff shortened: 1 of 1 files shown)


SinisterMinister:
This addresses #24. @victorgp Might consider changing from `su` to `exec sudo` which is what was [originally used](https://github.com/puppet-community/puppet-kafka/commit/6bd3c8fb5dc12a64aa4a26ef918ed8665206f1e3#diff-ea6062c201fe54c7a2b8e269a3a0b70bL33) before it broke.

victorgp:
@SinisterMinister that change was a good change intended to run kafka with kafka user, which is something good. We should keep that but removing "runuser", and the natural substitution for "runuser" is "su", so i prefer to use "su", that is known to work everywhere rather than coming back to "exec sudo"

SinisterMinister:
That makes sense and sounds good to me.

dhoppe:
Hello Victor, 

thank you very much for your contribution. This issue has been resolved by commit https://github.com/puppet-community/puppet-kafka/commit/389fd55d0e75b23e00aa4915babfb74a4bc5508a.

Kind regards,
Dennis

PR #25, opened 2015-11-14
Use su instead of runuser

sw0x2A:
runuser is not available on Ubuntu, su should be more compatible


dhoppe:
@sw0x2A, @stevie-: Thank you very much for your contribution. This issue has been resolved by commit ae994d5.

Kind regards,
Dennis

PR #29, opened 2015-12-14
Update module to version 1.0.3

dhoppe:


juniorsysadmin:
Just a few things:
I'm not sure I see any benefit in top scoping every variable here.
If you're removing a parameter, that according to Semantic versioning is a breaking change, and therefore there should be a major version bump (2.0)
And if you're feeling slightly bored, consider performing a modulesync with the puppet-community modulesync_configs - puppet-lint is likely not happy with the alignment of the variable blocks.

daenney:
Considering this module is no longer supporting Puppet 2.7 I have to agree with @juniorsysadmin. If this has been the style from the beginning sure but now it's just a lot of extra churn.

This should indeed also be a major version upgrade. You could leave the parameter in place and gracefully degrade but removing it outright needs a different version bump.

dhoppe:
Alright. I removed the changes regarding ```::``` and just changed the value of ```install_dir``` instead of removing the parameter. 

I will also take a look at modulesync_configs, because my current Lint setup does not complain about the alignment.

manifests/consumer.pp
```diff
@@ -56,7 +56,7 @@
 ) inherits kafka::params {
 
   validate_re($::osfamily, 'RedHat|Debian\b', "${::operatingsystem} not supported")
-  validate_absolute_path($install_dir)
+  #validate_absolute_path($install_dir)
```
daenney: Please remove this, don't uncomment it. Git keeps the history for us if we need to revert.

manifests/mirror.pp
```diff
@@ -61,7 +61,7 @@
 ) inherits kafka::params {
 
   validate_re($::osfamily, 'RedHat|Debian\b', "${::operatingsystem} not supported")
-  validate_absolute_path($install_dir)
+  #validate_absolute_path($install_dir)
```
daenney: Please remove this, don't uncomment it. Git keeps the history for us if we need to revert.

manifests/producer.pp
```diff
@@ -57,7 +57,7 @@
 ) inherits kafka::params {
 
   validate_re($::osfamily, 'RedHat|Debian\b', "${::operatingsystem} not supported")
-  validate_absolute_path($install_dir)
+  #validate_absolute_path($install_dir)
```
daenney: Please remove this, don't uncomment it. Git keeps the history for us if we need to revert.

dhoppe:
Done. I also removed the validations of ```$version``` and ```$scala_version```.

manifests/broker/topic.pp
```diff
@@ -17,7 +17,7 @@
   if $ensure == 'present' {
     exec { "create topic ${name}":
       command => "/opt/kafka/bin/kafka-topics.sh --create --zookeeper '${zookeeper}' --replication-factor ${replication_factor} --partitions ${partitions} --topic ${name}",
-      unless  => "/bin/bash -c \"if [[ \\\"`/opt/kafka/bin/kafka-topics.sh --list --zookeeper '${zookeeper}' ${name}`\\\" == *${name}* ]]; then exit 0; else exit 1; fi\""
+      unless  => "/opt/kafka/bin/kafka-topics.sh --list --zookeeper ${zookeeper} | grep '^${name}\$'",
```
igalic: why not just use `grep -x ${name}` here?

dhoppe:
I would love to fix this, but somebody removed me from the team modules/admin.

```
dhoppe@macbook ~/Projects/puppet/puppet-kafka (git)-[v1.0.3] % git push origin v1.0.3
ERROR: Permission to voxpupuli/puppet-kafka.git denied to dhoppe.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

daenney:
@dhoppe Yes, I did. That team is meant for people to be in to facilitate transferring of the module, not a blanket team to just be in and have admin access to everything. I've just given collaborators access, as should've been done from the start, so you can push again.

igalic:
ugh, i'm very very sorry about this, @dhoppe!

manifests/broker/config.pp
```diff
@@ -8,7 +8,6 @@
 # It manages the broker config files
 #
 class kafka::broker::config(
-  $install_dir = $kafka::broker::install_dir,
```
dhoppe: This change does not justifiy a major version upgrade, because this parameter is not used anywhere.

dhoppe:
I would like to see the this patch being merged, because there are still other issues left.

jyaworski:
Hello:

Can you rebase and/or squash?

dhoppe:
@jyaworski I am confused. Why do I need to perform a rebase/squash?

jyaworski:
@dhoppe it turns your 9 commits into one, and makes the PR cleaner.

dhoppe:
@jyaworski Done, but it still makes no sense. Before the rebase / squash every commit was well documented.

jyaworski:
Thanks for the contribution, @dhoppe!

PR #35, opened 2016-01-28
Add support for optional default file with environment vars

knumor:
Would you consider something like this to allow setting variables like KAFKA_HEAP_OPTS, LOG_DIR etc. used by the start script?


dbeckham:
@dhoppe, there is no current way to increase the amount of memory that Kafka is given unless you manually modify the init.d script in this module.  What needs to be done to knumor's change to be able to be merged into your project? Does the change just need to have conflicts resolved?

If not, how would you like to see Java heap memory managed in this module?  I have to modify it for my own needs to support this in some way, so I'd rather make a change that I can push back to the project, instead of just forking and moving on.

dhoppe:
@knumor You are absolutely right, but I am not sure if we should use one file to manage three different services. Furthermore we also should manage the file ```/etc/default/kafka``` somehow, for example an .erb template if a variable has an value.

bastelfreak:
I'm okay with this for now. People who want to manage the `/etc/default/kafka` file can still do this in their profile or we add the needed logic later to the kafka module. Thanks for the work @dbeckham!

flozano:
Hi, 
@jolivares and me really need this feature (in order to set CLASSPATH to an initial value, so that we can plug a JAR for metric reporting from an external directory). Any plans to make a release for the module soon?

PR #32, opened 2016-01-27
Fix empty PID file creation on failed daemon start

Project0:
When daemon cannot be started or quits to fast, then no process id can be fetched by ps and empty pid files are created.

PR #40, opened 2016-03-07
Fix empty PID file creation on failed daemon start

dhoppe:
Please merge this, because the issuer of #32 does not respond.


bastelfreak:
:+1:

PR #44, opened 2016-03-08
Add missing dependency

dhoppe:
Please merge.


fessyfoo:
We're trying to figure out why  puppet-zookeeper needed for the puppet-kafka module.  seems that though zookeeper is needed to run kafka,   it's not required to install kafka.

PR #46, opened 2016-03-10
Fix classes broker::consumer and broker::producer

dhoppe:
I am writing a couple of beaker tests and came across the following problems:

The generated Inits Scripts were broken because of
- missing equal signs
- empty parameters
- obsolete parameters
- some parameters are required
- wrong process names

The file resource did not notified the correct service and the config parameters for properties file or Init Script should be managed at the main class. 

I know that this would force a major release, but on the other hand this part of the Kafka module could not have worked before.


bastelfreak:
could you add a note to the README.md that a major release is required the next time?

dhoppe:
@bastelfreak I will do the next release myself, but I am not sure if I should do this before or after adding the support for Kafka 0.9.

manifests/broker/config.pp
```diff
@@ -23,12 +25,12 @@
   }
 
   file { '/opt/kafka/config/server.properties':
+    ensure  => present,
```
igalic: `ensure => 'file'`

manifests/consumer/config.pp
```diff
@@ -8,25 +8,25 @@
...
+  $consumer_config = deep_merge($config_defaults, $config)
 
   $config_notify = $service_restart ? {
-    true    => Service['kafka'],
+    true    => Service['kafka-consumer'],
     default => undef
   }
 
   file { "/opt/kafka/config/${name}.properties":
     ensure  => present,
```
igalic: also here

templates/consumer.init.erb
```diff
@@ -21,7 +21,7 @@ start() {
     else
         exec $DAEMON $DAEMON_OPTS >> /var/log/kafka/consumer.log 2>&1 &
         sleep 2
-        PID=`ps axww | grep -E '[k]afka-console-consumer' | awk '{print $1}'`
+        PID=`ps axww | grep -E '[k]afka.tools.ConsoleConsumer' | awk '{print $1}'`
```
igalic: what's wrong with `pgrep -f kafka.tools.ConsoleConsumer`?

templates/producer.init.erb
```diff
@@ -21,7 +21,7 @@ start() {
     else
         exec $DAEMON $DAEMON_OPTS >> /var/log/kafka/producer.log 2>&1 &
         sleep 2
-        PID=`ps axww | grep -E '[k]afka-console-producer' | awk '{print $1}'`
+        PID=`ps axww | grep -E '[k]afka.tools.ConsoleProducer' | awk '{print $1}'`
```
igalic: what's wrong with `pgrep -f kafka.tools.ConsoleConsumer`?

templates/server.properties.erb
```diff
@@ -2,8 +2,8 @@
 #
 # See: http://kafka.apache.org/documentation.html#brokerconfigs for more details
 <%- @server_config.sort.map do |k,v|
-  if v.to_s.strip.length != 0 %>
+    if v.to_s.strip.length != 0 -%>
 <%= k %>=<%= v %>
```
igalic: we can use `<%= k %>=<%= v %><%- unless v.to_s.strip.empty? -%>` and thus, we can get rid of the double end at the bottom.

or, we can use filter after before map

manifests/producer/config.pp
```diff
@@ -8,23 +8,25 @@
...
   $producer_config = deep_merge($kafka::params::producer_config_defaults, $config)
 
   $config_notify = $service_restart ? {
-    true    => Service['kafka'],
+    true    => Service['kafka-producer'],
     default => undef
   }
 
   file { '/opt/kafka/config/producer.properties':
     ensure  => present,
```
igalic: ensure => file

manifests/broker/config.pp
dhoppe: This is not necessary and we do not use it for our other modules. 
https://docs.puppetlabs.com/puppet/latest/reference/type.html#file-attribute-ensure

manifests/consumer/config.pp
dhoppe: This is not necessary and we do not use it for our other modules. 
https://docs.puppetlabs.com/puppet/latest/reference/type.html#file-attribute-ensure

manifests/producer/config.pp
dhoppe: This is not necessary and we do not use it for our other modules. 
https://docs.puppetlabs.com/puppet/latest/reference/type.html#file-attribute-ensure

templates/consumer.init.erb
dhoppe: Nothing and I have no idea if the regular expression is necessary.

templates/producer.init.erb
dhoppe: Nothing and I have no idea if the regular expression is necessary. 
BTW: You missed ```init.erb``` und ```mirror.init.erb``` ;)

templates/server.properties.erb
dhoppe: Unfortunately this syntax also requires a second end, but your solutions is looking much cleaner.

manifests/broker/config.pp
igalic: i thought puppet-lint would complain…

templates/consumer.init.erb
igalic: that's not so much a regular expression, as it simply assures that grep itself won't show up in the grepped output

pgrep does the same thing as the above, but it does so while not needing two pipes, and three forks of three different binaries.

templates/producer.init.erb
igalic: ;)

manifests/broker/config.pp
dhoppe: No that was about the broken class ```mirror```.

igalic:
nicely done.

are you ready for merge, or do you want to pretty up your commit history?

bastelfreak:
IMO the history is fine, nicely seperated small commits

dhoppe:
Give me some time to clear my head. In the meantime I will run some Beaker tests to make sure the configs and init scripts are still generated correctly. 

If this is working as expected, I will add the spec files for the Beaker tests as a separate  PR.

igalic:
:+1:

bastelfreak:
:+1:

PR #49, opened 2016-03-14
Submit service_name to cosumer::config and producer::config

dhoppe:
Please merge. This should fix #2.

### DIFF
--- a/manifests/mirror/config.pp
+++ b/manifests/mirror/config.pp
@@ -22,12 +22,14 @@ class kafka::mirror::config(
   ::kafka::consumer::config { 'consumer':
     config          => $consumer_config,
     config_defaults => $consumer_config_defaults,
+    service_name    => 'kafka-mirror',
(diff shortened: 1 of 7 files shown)


manifests/mirror/config.pp
```diff
@@ -19,15 +19,27 @@
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  ::kafka::consumer::config { 'consumer':
+  if $consumer_config['group.id'] == '' {
```
igalic: what will this do if the key `group.id` doesn't exist in `$consumer_config`?

dhoppe: It will throw an error message like ```fail('[Consumer] You need to specify a value for group.id')```. Otherwise the service ```kafka::mirror``` will not start and will complain about the missing parameter at ```/var/log/kafka/mirror.log```.

PR #50, opened 2016-03-14
Use module voxpupuli/archive instead of wget. This should fix #31

### LINKED ISSUE #31: stop trying to make fetch happen
stop trying to make fetch happen. it's not going to happen!

@voxpupuli has [puppet-archive](https://github.com/voxpupuli/puppet-archive/) which abstracts away this ad-hoc implementation of fetch, I mean wget. we should use that instead.

i realize that this is a breaking change, but perhaps it is for the better.

dhoppe:
Please merge.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,7 +1,8 @@
 fixtures:
   repositories:
-    zookeeper: "https://github.com/deric/puppet-zookeeper.git"
+    archive:   "https://github.com/voxpupuli/puppet-archive.git"
     java:      "https://github.com/puppetlabs/puppetlabs-java.git"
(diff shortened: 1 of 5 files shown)


igalic:
i was going to suggest adding checksum verification, but, what the heck is this checksum format??
https://dist.apache.org/repos/dist/release/kafka/0.9.0.1/

dhoppe:
I think this makes no sense, because there are multiple combinations of Kafka and Scala versions possible.

igalic:
we can build the checksum url the same way we build the download url (except we'd have to use dist.apache.org rather than a mirror)

dhoppe:
@igalic I am happy to merge, if you create a pull request. ;)

PR #55, opened 2016-03-15
Extend Beaker tests

dhoppe:
Please merge.

There are some errors regarding the service of Kafkra Broker which should be fixed with version 0.9.0.0.
https://issues.apache.org/jira/browse/KAFKA-1711


manifests/params.pp
```diff
@@ -164,8 +164,8 @@
     'batch-size'                 => '200',
     'broker-list'                => '',
     'compression-codec'          => '',
-    'key-serializer'             => 'kafka.serializer.StringEncoder',
-    'line-reader'                => 'kafka.tools.ConsoleProducer$LineMessageReader',
+    'key-serializer'             => 'kafka.serializer.DefaultEncoder',
+    'line-reader'                => 'kafka.tools.ConsoleProducer\$LineMessageReader',
```
bastelfreak: is the \ needed? the string is in single quotes

bastelfreak:
looks good besides the \

manifests/params.pp
dhoppe: Yes, this hash is passed through an .erb template. Otherwise the ```$LineMessageReader``` would be replaced with an whitespace.

PR #57, opened 2016-03-28
Support CentOS 7

bjoernhaeuser:
Adapt module to work better with CentOS7/RedHat 7

Also add configurable jmx/log4j/gc options to broker.

### DIFF
--- a/manifests/broker/service.pp
+++ b/manifests/broker/service.pp
@@ -26,13 +26,10 @@ class kafka::broker::service(
       content => template('kafka/init.erb'),
     }
 
-    if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
-      exec { 'systemctl daemon-reload # for kafka':
-        command     => '/bin/systemctl daemon-reload',
-        refreshonly => true,
-        subscribe   => File['/etc/init.d/kafka'],
-        before      => Service['kafka'],
-      }
+    if $::service_provider == 'systemd' {
+      include ::systemd
+
+      File['/etc/init.d/kafka'] ~> Exec['systemctl-daemon-reload'] -> Service['kafka']
     }
 
     service { 'kafka':
--- a/manifests/broker/service.pp
+++ b/manifests/broker/service.pp
@@ -9,7 +9,10 @@
 #
 class kafka::broker::service(
   $service_install = $kafka::broker::service_install,
-  $service_ensure  = $kafka::broker::service_ensure
+  $service_ensure  = $kafka::broker::service_ensure,
+  $jmx_opts        = $kafka::broker::jmx_opts,
+  $gc_opts         = $kafka::broker::gc_opts,
+  $log4j_opts      = $kafka::broker::log4j_opts,
 ) {
 
   if $caller_module_name != $module_name {
@@ -23,6 +26,15 @@ class kafka::broker::service(
       content => template('kafka/init.erb'),
     }
 
+    if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
+      exec { 'systemctl daemon-reload # for kafka':
+        command     => '/bin/systemctl daemon-reload',
+        refreshonly => true,
+        subscribe   => File['/etc/init.d/kafka'],
+        before      => Service['kafka'],
+      }
+    }
+
     service { 'kafka':
       ensure     => $service_ensure,
       enable     => true,
@@ -33,5 +45,4 @@ class kafka::broker::service(
   } else {
     debug('Skipping service install')
   }
-
 }
--- a/manifests/broker/service.pp
+++ b/manifests/broker/service.pp
@@ -20,16 +20,27 @@ class kafka::broker::service(
   }
 
   if $service_install {
-    file { '/etc/init.d/kafka':
-      ensure  => present,
-      mode    => '0755',
-      content => template('kafka/init.erb'),
-    }
-
     if $::service_provider == 'systemd' {
       include ::systemd
 
-      File['/etc/init.d/kafka'] ~> Exec['systemctl-daemon-reload'] -> Service['kafka']
+      file { '/usr/lib/systemd/system/kafka.service':
+        ensure  => present,
+        mode    => '0644',
+        content => template('kafka/broker.unit.erb'),
+      }
+
+      file { '/etc/init.d/kafka':
+        ensure => absent,
+      }
+
+      File['/usr/lib/systemd/system/kafka.service'] ~> Exec['systemctl-daemon-reload'] -> Service['kafka']
+    } else {
+      file { '/etc/init.d/kafka':
+        ensure  => present,
+        mode    => '0755',
+        content => template('kafka/init.erb'),
+        before  => Service['kafka'],
+      }
     }
 
     service { 'kafka':
@@ -37,7 +48,6 @@ class kafka::broker::service(
       enable     => true,
       hasstatus  => true,
       hasrestart => true,
-      require    => File['/etc/init.d/kafka'],
     }
   } else {
     debug('Skipping service install')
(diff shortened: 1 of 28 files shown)


manifests/mirror/service.pp
```diff
@@ -27,6 +27,15 @@
     content => template('kafka/mirror.init.erb'),
   }
 
+  if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
+    exec { 'systemctl daemon-reload # for kafka-mirror':
```
bastelfreak: defining this exec multiple times isn't a good idea

bastelfreak:
reloading systemd is always painful. I think there is a module on the forge that provides the exec with an "unless defined" wrapped around?m aybe we could use that instead?

manifests/mirror/service.pp
bjoernhaeuser: Its only defined once per mirror/producer/broker service.

bjoernhaeuser:
Do you have a hint of the module? I looked into other voxpupuli modules and they also define the same exec.

igalic:
must be one of these 11… https://forge.puppetlabs.com/modules?utf-8=%E2%9C%93&sort=rank&q=systemd

jyaworski:
@igalic if the service file changes, you have to issue a daemon-reload.

juniorsysadmin:
This one: https://forge.puppetlabs.com/camptocamp/systemd - but the namespace could have been named a little better

bjoernhaeuser:
I added the camptocamp/systemd module. Tests are running fine, now trying the acceptance tests :)

manifests/broker/service.pp
```diff
@@ -23,6 +26,12 @@
       content => template('kafka/init.erb'),
     }
 
+    if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
```
jyaworski: This would also be valid on Debian 8 and later Ubuntu. Can you use <code>$::serviceprovider<code> here?

bjoernhaeuser: Where is this fact defined? Cannot find it :-(

bjoernhaeuser: (And I actually dont like hardcoding these values here)

manifests/params.pp
```diff
@@ -312,7 +317,6 @@
     'broker-list'                => '',
     'compression-codec'          => '',
     'key-serializer'             => 'kafka.serializer.DefaultEncoder',
-    'line-reader'                => 'kafka.tools.ConsoleProducer$LineMessageReader',
```
bjoernhaeuser: This breaks the console producer in 0.9 :/

manifests/producer.pp
```diff
@@ -46,6 +46,7 @@
 # }
 #
 class kafka::producer (
+  $input,
```
bjoernhaeuser: The console producer makes only sense if there is the possibility to read from something

manifests/broker/service.pp
jyaworski: It's in stdlib:

https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/facter/service_provider.rb

bjoernhaeuser: in Debian 8 the service provider is debian m(

bjoernhaeuser: Okay, only for puppet 3 :)

bjoernhaeuser:
Using stdlib `$::service_provider` fact now

bastelfreak:
Could you also replace the legacy sysv init script by a systemd service file? I'm not sure if upstream ships one, but several distributions already do, for example [Arch Linux](https://aur.archlinux.org/cgit/aur.git/tree/systemd_kafka.service?h=kafka).

bjoernhaeuser:
@bastelfreak 

I added unit files for the broker-service. Could you please take a look and tell me if thats the right approach? If yes I would continue and provide unit files for the other services also.

manifests/producer/service.pp
```diff
@@ -31,6 +32,12 @@
     content => template('kafka/producer.init.erb'),
   }
 
+  if $::service_provider == 'systemd' {
+    include ::systemd
+
+    File['/etc/init.d/kafka-producer'] ~> Exec['systemctl-daemon-reload'] -> Service['kafka-producer']
```
bastelfreak: we need to take care here about the unit file and set the init file to absent.

spec/classes/producer_spec.rb
```diff
@@ -45,4 +47,40 @@
...
+
+    describe 'kafka::producer::config' do
+      context 'defaults' do
+        it { is_expected.to contain_file('/opt/kafka/config/producer.properties') }
+      end
+    end
+
+    describe 'kafka::producer::service' do
+      context 'defaults' do
+        it { is_expected.to contain_file('/etc/init.d/kafka-producer').that_notifies('Exec[systemctl-daemon-reload]') }
```
bastelfreak: same as above, adjust to unit file path.

templates/broker.unit.erb
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Apache Kafka server (broker)
+Documentation=http://kafka.apache.org/documentation.html
+Requires=network.target remote-fs.target
+After=network.target remote-fs.target
```
bastelfreak: remote-fs.target isn't needed here. Does it require zookeeper?

bjoernhaeuser: Removed the remote-fs. Zookeeper is needed, but it normally does not run on the same machine.

bjoernhaeuser:
Added all unit files \o/

manifests/broker/service.pp
```diff
@@ -17,10 +20,27 @@
...
-    file { '/etc/init.d/kafka':
-      ensure  => present,
-      mode    => '0755',
-      content => template('kafka/init.erb'),
+    if $::service_provider == 'systemd' {
+      include ::systemd
+
+      file { '/usr/lib/systemd/system/kafka.service':
+        ensure  => present,
+        mode    => '0755',
```
bastelfreak: any reason for that mode? default for unit files is 0644.

manifests/producer/service.pp
```diff
@@ -25,10 +26,15 @@
     fail('[Producer] You need to specify a value for topic')
   }
 
-  file { '/etc/init.d/kafka-producer':
-    ensure  => present,
-    mode    => '0755',
-    content => template('kafka/producer.init.erb'),
+  if $::service_provider == 'systemd' {
+    fail('Console Producer is not supported on systemd, because the stdin of the process cannot be redirected')
```
bastelfreak: I'm not exactly sure about the root of the issue. I think that should be possible. Can we discuss that in IRC?

templates/mirror.unit.erb
```diff
@@ -0,0 +1,18 @@
...
+Description=Apache Kafka server (mirror)
+Documentation=http://kafka.apache.org/documentation.html
+Requires=network.target
+After=network.target
+
+[Service]
+Type=simple
+User=kafka
+Group=kafka
+SyslogIdentifier=kafka-consumer
```
bastelfreak: same identifier as consumer.unit.erb, is this on purpose?

manifests/producer/service.pp
bastelfreak: will keep it for now. handling the producer with systemd is a bit dirty and in general rarely used. We can still implement this if it is requested

metadata.json
```diff
@@ -18,11 +18,15 @@
     },
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 3.0.0"
+      "version_requirement": ">= 4.10.0"
```
bastelfreak: you've got 4.11 in fixtures. Can you bump that version here too?

dhoppe:
@bjoernhaeuser Thank you very much for this great pull request.

PR #59, opened 2016-04-10
Adding support to KAFKA_HEAP_OPTS

ortz:
Hey,

We're missing the ability to modify KAFKA HEAP OPTS, I hope this PR would be sufficient to add this support.

Thanks,

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -59,6 +59,7 @@ class kafka::broker (
   $service_restart = $kafka::params::service_restart,
   $jmx_opts        = $kafka::params::broker_jmx_opts,
   $gc_opts         = $kafka::params::broker_gc_opts,
+  $heap_opts       = $kafka::params::broker_heap_opts,
   $log4j_opts      = $kafka::params::broker_log4j_opts,
 ) inherits kafka::params {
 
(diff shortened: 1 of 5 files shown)


bastelfreak:
Hi @ortz, thanks for the contribution. How does $heap_opts and $gc_opts differ? The both have the same defaults. Does kafka handles these environment variables differently?

ortz:
Yes it does, if you'll take a look at kafka-start.sh it checks this env
variable and applies it if exists.

bastelfreak:
Alright. Could you also add tests for the new variable please?

ortz:
Yes, I'll add it tomorrow when I'll be at the office

ortz:
Hey @bastelfreak, I've added the tests.
What are the chances you'll accept the pull request and release a new version today ?
(In my project I also need CentOS7 support but it's missing in the latest release, it is merged to master though)

bjoernhaeuser:
Hey,

isnt that handled through GC opts?

ortz:
No, it doesn't look like that (https://apache.googlesource.com/kafka/+/trunk/bin/kafka-server-start.sh).

I couldn't find GC_OPTS in the server start file and HEAP_OPTS is been used.

bjoernhaeuser:
But then we need to remove the gc opts when they are not set.

DjxDeaf:
@bjoernhaeuser maybe you meant "KAFKA_GC_LOG_OPTS" which is being used in the bin/kafka-run-class.sh but leaving it empty is just fine, and setting the other params are a not related and quite important

bastelfreak:
thanks for the contribution!

PR #60, opened 2016-04-11
Remove archive include

ortz:
There is no need to include archive because it's a define and not a class.
This breaks the puppet (no such module ::archive).
You can take a look at https://github.com/camptocamp/puppet-archive and see there is no need to include archive.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -118,8 +118,6 @@ class kafka (
     ],
   }
 
-  include '::archive'
-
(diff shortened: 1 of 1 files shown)


ortz:
Hey @bastelfreak, here is another fix :)
I'd appreciate if you'll merge both of the pull requests and create a release.

ortz:
It seems the archive is messed up (wrong params and such), I'm checking.

ortz:
Ignore it, different archive modules.

PR #65, opened 2016-04-24
add the ability to optimize all the different roles, not just the broker

DjxDeaf:
Hi,

This module lacks the ability to, for example, set a different JMX (or at all) for the kafka mirrormaker, or set a different log4j config, etc.

This commit includes adding 3 additional sets of params, which all derives from the broker config, unless you set them - which is something you can do as all the relevant manifests are now able to receive them as variables.

By default the JMX ports are:
Broker - 9990 (No change, same as before)
Mirror - 9991
Producer - 9992
Consumer - 9993

I have encountered a situration where I have to install several components on the same host, which is one of things this commit tries to solve.

What do you think?

Thanks

### DIFF
--- a/manifests/consumer.pp
+++ b/manifests/consumer.pp
@@ -45,17 +45,19 @@
 #  config => { 'client.id' => '0', 'zookeeper.connect' => 'localhost:2181' }
 # }
 class kafka::consumer (
-  $version          = $kafka::params::version,
-  $scala_version    = $kafka::params::scala_version,
-  $install_dir      = $kafka::params::install_dir,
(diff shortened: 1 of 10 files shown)


jyaworski:
@DjxDeaf this has merge conflicts. Can you rebase?

DjxDeaf:
@jyaworski done, also removed the gc_opts per to the latest merged PR

bastelfreak:
Hi, thanks for the contribution!
Could you please also add tests for the new variables? Then I'm happy to merge.

DjxDeaf:
@bastelfreak Done

PR #66, opened 2016-04-24
Removing broker gc opts

bjoernhaeuser:
The variable name was wrong, it should be heap opts.

In #59 we figured out that I added the wrong env variable. #59 adds the right one, this removes the wrong one.


bjoernhaeuser:
\o/ thank you ruby

PR #67, opened 2016-04-26
add the jmx_opts and log4j_opts to the init scripts

DjxDeaf:
they were only added to the systemd scripts, now it should support both


bastelfreak:
thx for the contribution!

bastelfreak:
could you also add tests for the changes please? Then I'm happy to merge.

DjxDeaf:
they are already inside from the previous PR

PR #70, opened 2016-05-13
Fix wrong syntax (default value for ssl.enabled.protocols)

jolivares:
Despite the oficial Kafka doc says default value for `ssl.enabled.protocols`is `[TLSv1.2, TLSv1.1, TLSv1]`. It's wrong, and you'll get the following error if you use that syntax:
```java
Caused by: java.lang.IllegalArgumentException: [TLSv1.2
	at sun.security.ssl.ProtocolVersion.valueOf(ProtocolVersion.java:187)
	at sun.security.ssl.ProtocolList.convert(ProtocolList.java:84)
	at sun.security.ssl.ProtocolList.<init>(ProtocolList.java:52)
	at sun.security.ssl.SSLEngineImpl.setEnabledProtocols(SSLEngineImpl.java:2081)
	at org.apache.kafka.common.security.ssl.SslFactory.createSslEngine(SslFactory.java:138)
```
As found in other parts of the documentation it has to be set w/o the brackets:
```
ssl.enabled.protocols=TLSv1.2,TLSv1.1,TLSv1 (list out the SSL protocols that you are going to accept from clients. Do note that SSL is deprecated in favor of TLS and using SSL in production is not recommended)
```


manifests/params.pp
```diff
@@ -146,7 +146,7 @@
     'security.inter.broker.protocol'                => 'PLAINTEXT',
     'ssl.cipher.suites'                             => '',
     'ssl.client.auth'                               => 'none',
-    'ssl.enabled.protocols'                         => '[TLSv1.2, TLSv1.1, TLSv1]',
+    'ssl.enabled.protocols'                         => 'TLSv1.2, TLSv1.1, TLSv1',
```
jyaworski: Could you instead have an array, then call .join(',') on it?

jolivares:
@jyaworski I did some modifications to use an array.

PR #74, opened 2016-06-08
Allowing installation via package

Mike-Petersen:
Hi,
we are installing kafka from rpm. So i added the possibility to install via package.

Regards
Mike

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -35,12 +35,14 @@
 #
 #
 class kafka (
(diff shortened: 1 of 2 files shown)


juniorsysadmin:
Just a comment: ```version``` on its own generally indicates that specifying the package version is a required parameter value but with ```package_ensure``` it is not. We've got both now.

Mike-Petersen:
Thanks

PR #75, opened 2016-06-12
Fix HEAP_OPTS in init template

jcochard:


bastelfreak:
Hi @jcochard, thanks for the contribution \o/
Could you add a spec test for the change?

igalic:
this is the same as #83 and was also fixed in #82.
however, i didn't add any tests for this either ~_~

PR #76, opened 2016-06-16
Removed dependency which is not valid with package installation

Mike-Petersen:
Sorry, i missed to commit the require for the package installation to work.
Is that require in any way necessary for the archive?
Otherwise i would just remove it.


manifests/init.pp
```diff
@@ -104,10 +104,9 @@
   }
 
   file { '/opt/kafka/config':
-    ensure  => directory,
-    owner   => 'kafka',
-    group   => 'kafka',
-    require => Archive["${package_dir}/${basefilename}"],
```
igalic: perhaps we should put this dependency elsewhere then?

Mike-Petersen: I now added the dependency to the archive/package

Mike-Petersen:
I have retriggered the build cause it was an bundle error but it seems that there is now a problem with the new puppet 4.6 version (10.08.16). This error is not due to the code.

The problem is:
if versioncmp($kafka::version, '0.9.0.0') < 0 {

which causes an "Unknown variable: 'kafka::version'" error since puppet 4.6

Does anyone know how to fix this?

bastelfreak:
I noticed the same yesterday, a lot of stuff is currently breaking because of puppet4. Can you please pin puppet to ~> 4.5.0 in the Gemfile? Then we can see if the tests are working.

Mike-Petersen:
I also updated the .travis.yml to use puppet ~> 4.5.0

fessyfoo:
this totally makes sense.

PR #80, opened 2016-06-30
Ubuntu 16.04 systemd fix

nicobn:
Added tweak for systemd Ubuntu support. Tested on Ubuntu 16.04. By default, the /usr/lib/systemd/system directory does not exist on Ubuntu. It appears the prefered method of adding a service is by adding the unit definition in /etc/systemd/system.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,95 @@
+This module has grown over time based on a range of contributions from
+people using it. If you follow these contributing guidelines your patch
+will likely make it into a release a little quicker.
+
+
+## Contributing
+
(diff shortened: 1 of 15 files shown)


bastelfreak:
/usr/lib/systemd/system and /usr/lib/systemd/system are the paths used for unit files provided by the package manager. /etc/systemd is for user provided stuff. We should _always_ write to /etc. changing that know would be a breaking change.. :(

igalic:
imo, this whole mess should be hidden away in params.pp

nicobn:
@bastelfreak one imperfect way to avoid BC would be to ensure that the `/usr/lib/systemd/system` file is absent in the manifest.

PR #82, opened 2016-07-04
init clean-up

igalic:
This WIP pull request tries to clean-up our init scripts.
They are currently quadrupled (broker, consumer, mirror, producer). Unnecessarily so.

This pr will first attempt to clean up the shell scripts, then consolidate them into one template.


templates/consumer.init.erb
```diff
@@ -21,29 +34,29 @@ if [ -f /etc/default/kafka ]; then
 fi
 
 start() {
-  if [ -f $PID_FILE ]
-    then
+  if [ -f $PID_FILE ]; then
```
bastelfreak: this needs quotes around the var

templates/consumer.init.erb
```diff
@@ -21,29 +34,29 @@ if [ -f /etc/default/kafka ]; then
...
       PID=`cat $PID_FILE`
-      if [ -z "`ps -ef | awk '{print $2}' | grep "^$PID$"`" ]; then
+      if `ps -p "$PID" -o pid= ` -eq `pgrep -f "$PGREP_PATTERN"`; then
```
bastelfreak: can you migrate to the new bash style $()?

templates/consumer.init.erb
```diff
@@ -21,29 +34,29 @@ if [ -f /etc/default/kafka ]; then
...
-        fi
-    fi
+
+      echo "\n$PID_FILE exists but the process is not running. Deleting $PID_FILE and re-trying"
+      rm -f $PID_FILE
+      start
+      return $?
+  fi
+
+  /bin/su $KAFKA_USER -c "$DAEMON $DAEMON_OPTS >> /var/log/kafka/consumer.log 2>&1 &"
```
bastelfreak: please add quotes around $KAFKA_USER

templates/consumer.init.erb
```diff
@@ -21,29 +34,29 @@ if [ -f /etc/default/kafka ]; then
...
+  sleep 2
+  PID=`pgrep -f "$PGREP_PATTERN"`
```
bastelfreak: please migrate to the new bash style $()

templates/consumer.init.erb
```diff
@@ -21,29 +34,29 @@ if [ -f /etc/default/kafka ]; then
...
+  if [ -z "$PID" ]; then
+    echo "\n $NAME could not be started"
+    exit 1
+  fi
+
+  echo $PID > $PID_FILE;
```
bastelfreak: please add quotes around the vars

templates/consumer.init.erb
```diff
@@ -60,21 +73,19 @@ stop() {
...
-    if [ -z "`ps -ef | awk '{print $2}' | grep "^$PID$"`" ]
-    then
-      echo "$NAME stopped but pid file exists"
-      exit 1
-    else
-      echo "$NAME running with pid $PID"
-      exit 0
-    fi
-  else
+  if [ ! -f $PID_FILE ]; then
```
bastelfreak: please add quotes around the var

templates/consumer.init.erb
igalic: it's written in sh. not sure i can

igalic:
thank you for the review, @bastelfreak.
i'll add proper quoting once i've merged all three files into one :)
as far as `` vs $() goes: i'm not sure i can do this. debuntu's systemv prescribes `/bin/sh` (i don't know about rhel)

bastelfreak:
on a first look this looks good. Are we sure that zookeeper.service always has that name or does it need to be configurable?
I will have to review this when I've a few hours of free time, or a day.

igalic:
i'm contemplating to review it by putting it into production.

bastelfreak:
meme

igalic:
close/open did nothing to make @voxpupuli/pcci pick it up

templates/init.erb
```diff
@@ -27,77 +69,73 @@ start() {
...
   ulimit -s 10240
   ulimit -c unlimited
-  if [ -f $PID_FILE ]
-  then
-    PID=`cat $PID_FILE`
-    if [ ! -z "`ps -ef | awk '{print $2}' | grep "^$PID$"`" ]
-    then
+  if [ -f "$PID_FILE" ]; then
+    PID=`cat "$PID_FILE"`
+    if [ `ps -p "$PID" -o pid= || echo 1` -eq `pgrep -f "$PGREP_PATTERN" || echo 2` ] ; then
```
igalic: i really hope this isn't too clever

templates/init.erb
```diff
@@ -27,77 +69,73 @@ start() {
...
-      exit 0
-    fi
-  else
+  if [ ! -f "$PID_FILE" ]; then
     echo "$NAME stopped"
     exit 1
   fi
+
+  PID=`cat "$PID_FILE"`
+  if [ `ps -p "$PID" -o pid= || echo 1` -eq `pgrep -f "$PGREP_PATTERN" || echo 2` ] ; then
```
igalic: again, i hope this isn't too clever…

igalic:
it took creating and destroying 3 stacks, but with the latest patches it now works.

templates/init.erb
bastelfreak: that looks really strange. Have you tested that?

manifests/broker/service.pp
```diff
@@ -12,38 +12,46 @@
...
+
   if $service_install {
     if $::service_provider == 'systemd' {
       include ::systemd
 
-      file { '/usr/lib/systemd/system/kafka.service':
-        ensure  => present,
+      file { "${service_name}.service":
+        ensure  => file,
+        path    => "/usr/lib/systemd/system/${service_name}.service",
```
bastelfreak: overwriting anything in /usr/lib/systemd/system is a bad idea and not recommened. The prefered solution is to use /etc/systemd/system. The first path is meant to be used by package managers only. You can create a unit at the second path with an equal name which will kind of inherit from the first one. Also ubuntu or debian don't use /usr/lib but /lib/something.

templates/unit.erb
```diff
@@ -0,0 +1,38 @@
+[Unit]
+Description=Apache Kafka server (<%= (@service_name.split(/-/)[1] and @service_name.split(/-/)[1].capitalize) or 'broker' -%>)
+Documentation=http://kafka.apache.org/documentation.html
+Wants=zookeeper.service
+After=zookeeper.service
+Requires=zookeeper.service
```
bastelfreak: does the name of the zookeeper service needs to be configurable?

manifests/broker/service.pp
igalic: so, `/etc/systemd/system` ?

templates/unit.erb
igalic: most probably. but for now that's what our upstream module calls it: https://github.com/deric/puppet-zookeeper/blob/master/manifests/service.pp#L11

n.b.: our upstream module does the same bad thing to systemd.

manifests/broker/service.pp
bastelfreak: yep. I'm currently not sure how exactly the overwrites work. Not everything is possible here. I've to read up the docs.
https://wiki.archlinux.org/index.php/systemd
https://www.freedesktop.org/software/systemd/man/systemd.unit.html
https://www.freedesktop.org/software/systemd/man/systemd.service.html#

templates/unit.erb
bastelfreak: please stop reminding me :D

maybe we should define a policy/guideline somewhere which mentions a way to handle service files. So we can make it consistent across our modules (in a far far future)

templates/init.erb
igalic: yes!! it works really well

what it does is this:
if `ps -p 2194 -o pid=` doesn't find that pid to be running, it'll **output** nothing, but **return** 1. In that case the `|| echo 1` case will fire. simiarily, if `pgrep -f kafka.Kafka` doesn't find anything, it will **output** nothing, but **return** 1. Now, `|| echo 2` is the result. hence, we're comparing
`[ 1 -eq 2 ]`, which will be false. but! if one of `ps` or `pgrep` returns a sensible value, something is off! and we're catching those cases, too: `[ 2967 -eq 2 ]` is just as false, as `[ 1 -eq 2957 ]` is.

bastelfreak:
two fails on travis, this is based on a gem fuckup. the rest is fine.

templates/unit.erb
igalic: or we could use something like pleaserun…

PR #83, opened 2016-07-04
Parameter `heap_opts`  is missing from kafka::broker::service class

nicobn:
The variable is used in the `kafka/broker.unit.erb` template but is not assigned in the kafka::broker::service class. As such, it is not propagated from the kafka::broker class, unlike $kafka::broker::opts or $kafka::broker::jmx_opts.


igalic:
hi @nicobn !
did you notice #82? i'm working on cleaning up our init scripts and it's progressing nicely. i am going to merge your change there, if you think that's okay. if you rather it's merged *now* — and we release a fresh version with your changes before merging #82, that's okay too.

igalic:
this was fixed in #82
thanks for pointing it out, @nicobn!

nicobn:
Thank you !

PR #84, opened 2016-07-04
Added tweak for systemd Ubuntu support

nicobn:
Tested on Ubuntu 16.04. By default, the `/usr/lib/systemd/system` directory does not exist on Ubuntu. It appears the prefered method of adding a service is by adding the unit definition in `/etc/systemd/system`.

See #80 (closed by mistake - sorry about that) for discussion.


igalic:
this was fixed in #82

nicobn:
Thank you !

PR #87, opened 2016-07-05
modulesync 0.9.0

bastelfreak:


bastelfreak:
@igalic I think we should merge as it. we havn't decided yet how we want to deal with rest client yet. this should be handled in a further modulesync.

igalic:
ok!

PR #94, opened 2016-08-15
Init fix

igalic:
currently our system-v init file's `status()` has flawed logic.
This pr tries to address that.

it also makes `stop()` more robust

### DIFF
--- a/templates/init.erb
+++ b/templates/init.erb
@@ -95,9 +95,15 @@ start() {
 }
 
 stop() {
-  if [ ! -f "$PID_FILE" ]; then
-    echo "$PID_FILE does not exist, process is not running"
-    return 1
+  if ! [ -f "$PID_FILE" ]; then
+    echo -n "$PID_FILE does not exist"
+    if PID=`pgrep -f "$PGREP_PATTERN"` ; then
+      echo -n ", but process is running"
+      echo "$PID" > "$PID_FILE"
+    else
+      echo -n ", and process is not running"
+      return 1
+    fi
   fi
 
   kill `cat "$PID_FILE"`;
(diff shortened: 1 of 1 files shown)


igalic:
we really need monads in bash…
and a better different vector for returning status and success, or… etc…

bastelfreak:
@igalic can you take a look at https://github.com/voxpupuli/puppet-kafka/pull/96 please?

fessyfoo:
actually check #95  but I think it's a less comprehensive version of this.

templates/init.erb
```diff
@@ -107,13 +113,13 @@ stop() {
 }
 
 status() {
-  if [ ! -f "$PID_FILE" ]; then
+  if ! [ -f "$PID_FILE" ]; then
```
fessyfoo: FWIW I don't think this matters.   `[` is equiv to the builtin `test`  and from the [posix description of test](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html):

> !  expression
True if expression is false. False if expression is true.

```
$ [ -f /doesnotexist ] && echo true || echo false
false
$ [ ! -f /doesnotexist ] && echo true || echo false 
true
```

I'm curious if you just did this for style purposes, habit, or some other need that I'm missing?

igalic: it's a mix of the above.

igalic: to me, `if ! [` denotes `unless [`, whereas `if [ !` denotes: ATTENTION: YOU HAVE TO THINK ABOUT LOGIC NOW

fessyfoo:
FWIW I recommend merging this.  without it the service is broken, and puppet will continually try to restart the service thinking it's down.

fessyfoo:
turns out I'm a committer on this project.   why are the checks failing?  ok, fixed that.

igalic:
@fessyfoo would you like to become a committer on all @voxpupuli projects?

PR #95, opened 2016-08-20
fix init script to actually check status

fessyfoo:
flip sense of the check that pid in pid file matches pid in process
list.

### DIFF
--- a/templates/init.erb
+++ b/templates/init.erb
@@ -113,7 +113,7 @@ status() {
   fi
 
   PID=`cat "$PID_FILE"`
-  if [ `ps -p "$PID" -o pid= || echo 1` -eq `pgrep -f "$PGREP_PATTERN" || echo 2` ] ; then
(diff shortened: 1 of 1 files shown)


igalic:
is this the same as #96 ?

fessyfoo:
#96 was  a merge to a different branch.     there was some confusion around opentable/puppet-kafka being renamed to voxpupuli/puppet-kafka  and the magic redirecting github does.

fessyfoo:
this is  a less comprehensive version of #94

fessyfoo:
FWIW we reforked opentable/puppet-kafka   so we have our own fork again.   [ voxpupli was a transfer owner ship of opentable/puppet-kafka and when the branch `no-zookeper-dep` was pushed to opentable/puppet-kafka  it showed up on voxpupli/puppet-kafka ....

blah hope that makes sense,  sorry for adding confusion.

PR #96, opened 2016-08-20
fix init script to actually check status

fessyfoo:
flip sense of the check that pid in pid file matches pid in process
list.


igalic:
:100:

Issue #98, opened 2016-08-26
Incompatibility beetween $install_dir and $version params


RNiveau:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.7
- Ruby: 2.0.0
- Distribution: CentOs7
- Module version: 2.0.0

### How to reproduce (e.g Puppet code you use)

```
  class { '::kafka' :
    install_java => false,
    version      => '0.10.0.1',
    install_dir  => '/mydir/kafka'
  }
```
### What are you seeing

Install directory is /opt/kafka-2.11-0.10.0.1

### What behaviour did you expect instead

Install directory should be /mydir/kafka

### Any additional information you'd like to impart

I think the bug comes from this piece of code (from init.pp):

```
if $version != $kafka::params::version {
    $install_directory = "/opt/kafka-${scala_version}-${version}"
  } elsif $scala_version != $kafka::params::scala_version {
    $install_directory = "/opt/kafka-${scala_version}-${version}"
  } else {
    $install_directory = $install_dir
  }
```

igalic:
i believe this to be fixed with #99

PR #99, opened 2016-08-30
Fix incompatibility beetween $install_dir and $version params

igalic:
This addresses #98.

if install_dir was *not* changed, we adapt it for the potentially supplied
scala_version and kafka version

if it *was* changed, we set it, as-is.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -54,12 +54,12 @@ class kafka (
   $basefilename = "kafka_${scala_version}-${version}.tgz"
   $package_url = "${mirror_url}/kafka/${version}/${basefilename}"
 
-  if $version != $kafka::params::version {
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -54,12 +54,12 @@
   $basefilename = "kafka_${scala_version}-${version}.tgz"
   $package_url = "${mirror_url}/kafka/${version}/${basefilename}"
 
-  if $version != $kafka::params::version {
-    $install_directory = "/opt/kafka-${scala_version}-${version}"
-  } elsif $scala_version != $kafka::params::scala_version {
-    $install_directory = "/opt/kafka-${scala_version}-${version}"
-  } else {
-    $install_directory = $install_dir
+  $install_directory = $install_dir ? {
```
bastelfreak: :+1: for selectors

PR #105, opened 2016-10-08
Make the systemd manifest configurable

bjoernhaeuser:
Zookeeper sometimes does not run on the same machine as kafka, having a
hard Requires=zookeeper.service in the unit file, will make kafka not
start in these cases.

Therfore adding a parameter, to disable this Requires directive.
bastelfreak mentioned that the default could change in the next major
release.


templates/unit.erb
```diff
@@ -1,9 +1,11 @@
 [Unit]
 Description=Apache Kafka server (<%= (@service_name.split(/-/)[1] and @service_name.split(/-/)[1].capitalize) or 'broker' -%>)
 Documentation=http://kafka.apache.org/documentation.html
+<% if @service_requires_zookeeper: %>
```
bastelfreak: the : shouldn't be here.

bjoernhaeuser: fixed

manifests/mirror/service.pp
```diff
@@ -8,14 +8,16 @@
...
 class kafka::mirror::service(
-  $consumer_config       = $kafka::params::consumer_config,
-  $producer_config       = $kafka::params::producer_config,
-  $num_streams           = $kafka::mirror::num_streams,
-  $num_producers         = $kafka::mirror::num_producers,
-  $abort_on_send_failure = $kafka::mirror::abort_on_send_failure,
-  $whitelist             = $kafka::mirror::whitelist,
-  $blacklist             = $kafka::mirror::blacklist,
-  $max_heap              = $kafka::mirror::max_heap
+  $service_ensure             = $kafka::mirror::service_ensure,
```
bastelfreak: I think the introduction to of $service_ensure isn't related to this bug? Can you move it to a separate commit?

templates/unit.erb
```diff
@@ -1,9 +1,11 @@
 [Unit]
 Description=Apache Kafka server (<%= (@service_name.split(/-/)[1] and @service_name.split(/-/)[1].capitalize) or 'broker' -%>)
 Documentation=http://kafka.apache.org/documentation.html
+<% if @service_requires_zookeeper %>
```
bastelfreak: please add spec tests that ensure zookeeper is/isn't present for producer,mirror,consumer,broker

bastelfreak:
We probably should also fix the hard zookeeper dependency in the legacy startup script https://github.com/voxpupuli/puppet-kafka/blob/master/templates/init.erb#L8

templates/unit.erb
bjoernhaeuser: Added, but not for producer, because this one does not work with systemd anyway

bastelfreak: I think we need -%> here? Or <%- ?

PR #108, opened 2016-10-13
adding the port to the mirror URL

petetodo:
in some organizations where external repositories are mirrored via maven/nexus the internal mirror URL will contain the port . Right now the validation on the mirror url fails because of that. adding just minor change to the reg exp to accept the port

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '0.13.3'
+modulesync_config_version: '0.14.1'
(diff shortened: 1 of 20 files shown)


bastelfreak:
Hi @petetodo, could you add spec tests to validate a few different URLs?

petetodo:
yeah will do : have to learn fast how to do it . but no worries will happen soon. 
I used https://regex101.com/ to validate the string so it has to be working :)

bastelfreak:
You can join us any time in #voxpupli on freenode if you've questions about spec tests or anything related.

PR #112, opened 2016-11-01
Update readme with the current Kafka & Scala versions

atrepca:
This is to Update README.md to reflect the versions of Kafka and Scala being installed.


dhoppe:
@atrepca Thank you very much for your contribution. I will ignore the results of Travis CI, because this are just changes regarding the documentation.

PR #116, opened 2016-11-08
Fix bounding of puppet version in metadata

ghoneycutt:
`pe` is no longer needed and is determined based on `puppet`.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -62,14 +62,7 @@
     }
   ],
   "requirements": [
-    {
-      "name": "pe",
-      "version_requirement": ">= 3.2.0"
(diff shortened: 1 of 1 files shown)


dhoppe:
In this case we should fix the requirements at https://github.com/voxpupuli/modulesync_config.

ghoneycutt:
@dhoppe I don't see where the metadata is managed by modulesync.

dhoppe:
@ghoneycutt Ups, I mixed this up with the Puppet skeleton.

ghoneycutt:
Thanks @dhoppe

Issue #117, opened 2016-11-18
broker.id should also be set in $logs_dir/meta.properties


igalic:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: any
- Ruby: any
- Distribution: any
- Module version: <= 2.1.0

### What are you seeing

When `kafka::broker::config`'s `broker.id` changes that change is not reflected in `$log_dir`'s `meta.properties`. This causes kafka startup to fail (see below)

### What behaviour did you expect instead

Dynamic environments are… well… dynamic. An integer such as a `broker.id` can easily be changed for any number of reasons. Our module should enable and support this.

### Output log

```
[2016-11-18 13:25:28,078] INFO Logs loading complete. (kafka.log.LogManager)
[2016-11-18 13:25:28,130] INFO Starting log cleanup with a period of 300000 ms. (kafka.log.LogManager)
[2016-11-18 13:25:28,131] INFO Starting log flusher with a default period of 9223372036854775807 ms. (kafka.log.LogManager)
[2016-11-18 13:25:28,138] FATAL Fatal error during KafkaServer startup. Prepare to shutdown (kafka.server.KafkaServer)
kafka.common.InconsistentBrokerIdException: Configured broker.id 3 doesn't match stored broker.id 1 in meta.properties. If you moved your data, make sure your configured broker.id matches. If you intend to create a new broker, you should remove all data in your data directories (log.dirs).
        at kafka.server.KafkaServer.getBrokerId(KafkaServer.scala:648)
        at kafka.server.KafkaServer.startup(KafkaServer.scala:187)
        at kafka.server.KafkaServerStartable.startup(KafkaServerStartable.scala:37)
        at kafka.Kafka$.main(Kafka.scala:67)
        at kafka.Kafka.main(Kafka.scala)
[2016-11-18 13:25:28,140] INFO shutting down (kafka.server.KafkaServer)
```
### Any additional information you'd like to impart

I don't know how to guarantee the uniqueness of this number across a cluster, since in puppet we have no easy way to know who's part of a cluster.

Zookeeper would be a way? maybe??

dhoppe:
@igalic The `broker.id` should never change and somehow you need to make sure that it is unique. In my case I use an AWS EC2 tag as fact to provide a unique `broker.id`. 

I do not know what you mean by _dynamic environments_, because I have not seen an environment where you scale up/out Kafka nodes. 

If you need this you might want to take a look at AWS Kinesis Firehose.

[...]

PR #118, opened 2016-11-21
WIP: change broker.id in meta.properties

igalic:
This change addresses #117: when `broker.id` changes in
`server.properties`, this should also be reflected in
`$log.dir/meta.properties`. This change will currently allow the broker
to startup, should those values have gotten out of sync for which ever
reason.
However, this will *not* match up leaders/replicas in zookeeper!

### DIFF
--- a/manifests/broker/config.pp
+++ b/manifests/broker/config.pp
@@ -38,12 +38,21 @@ class kafka::broker::config(
   }
 
   file { '/opt/kafka/config/server.properties':
-    ensure  => present,
+    ensure  => file,
     owner   => 'kafka',
     group   => 'kafka',
     mode    => '0644',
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hai, can you please add spec tests as well?

bastelfreak:
@igalic ping

igalic:
as pointed out by @dhoppe, https://github.com/voxpupuli/puppet-kafka/issues/117#issuecomment-262947854 this is probably sort of bad idea

igalic:
as pointed out by @dhoppe, https://github.com/voxpupuli/puppet-kafka/issues/117#issuecomment-262947854 this is probably sort of bad idea

PR #121, opened 2016-12-06
Mirror url custom port

ellamdav:
- This change makes validation of mirror_url consistent across broker/producer/consumer/mirror classes, allowing any of these to include a custom port#

- It also fixes an error in the optional /path/to/package part of the original URL regex, which effectively caused the *.tld 2-6 character length restriction to be ignored.

- Tests for the various elements of the URL regex are run combinatorially – ie. every possible combination is tested.  Since these are repeated for each of the 4 dependent classes, running all unit tests takes ~5mins.  Testing a few specific sample URLs would be quicker, but risks missing some edge cases.  Really this highlights that this common validation should be extracted into a custom puppet function, unit tested once there, and mocked out in tests of dependent classes.  That’s beyond the scope of this change though.

### DIFF
--- a/spec/classes/broker_spec.rb
+++ b/spec/classes/broker_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'shared_examples_param_validation'
 
 describe 'kafka::broker', type: :class do
   let :facts do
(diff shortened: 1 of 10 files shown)


spec/classes/broker_spec.rb
```diff
@@ -11,14 +12,18 @@
       service_provider: 'upstart'
     }
   end
-  let :params do
+  let :common_params do
```
bastelfreak: :+1:

bastelfreak:
thanks!

PR #125, opened 2016-12-26
Bump dependencies

juniorsysadmin:
- Don't restrict the range for camptocamp/systemd
- Set minimum version for puppet/archive to 1.0.0

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppet/archive",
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "camptocamp/systemd",
```
juniorsysadmin: Doesn't explicitly support Puppet 4 in metadata

PR #135, opened 2017-03-10
Ensure the process is killed before finishing stop

elmendalerenda:
There are problems when restarting the Kafka service, it seems that the process is not effectively killed before the start phase is initiated.
I saw many checks in the start method but I think we should block the start method until Kafka is fully killed.

### DIFF
--- a/templates/init.erb
+++ b/templates/init.erb
@@ -90,12 +90,20 @@ stop() {
     fi
   fi
 
-  kill `cat "$PID_FILE"`;
+  PID=`cat $PID_FILE`
+  kill $PID;
   rm -f -- "$PID_FILE";
 
   # wait until the process is finished
+  RETRIES=0
+  MAX_RETRIES=10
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @elmendalerenda, thanks for the PR. Will this lead to an infinity loop in some situations? is there any upstream init script that we could use here?

elmendalerenda:
You are right, I've implemented an strategy to exit the loop after a number of retries. I've been looking for a standard init script, but I haven't found anything like that (I'm not an expert, though).

elmendalerenda:
@bastelfreak is there any other concern with the PR?

bastelfreak:
Hi @elmendalerenda, sorry for the slow response. Can you please rebase against our current master and check if travis still throws errors? Also please check the used email address in the commits. It isn't associated with your github account.

elmendalerenda:
@bastelfreak there it  is. Changed author, rebased and green ci :tada:

bastelfreak:
thanks!

PR #137, opened 2017-03-16
Add params to allow customizing user, group, and log_dir

seanmalloy:

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,22 @@ Create kafka group with this ID
 
 Create kafka user with this ID
 
+#### `user`
+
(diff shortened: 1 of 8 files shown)


seanmalloy:
I am in the process of fixing the rspec-puppet tests. It looks like the puppet-archive module is now using structured facts which is causing the rspec-puppet tests for this module to fail. I'm hoping to fix all of the failing test cases in the next few days.

seanmalloy:
The rspec-puppet tests are now fixed. Let me know your thoughts on my proposed code changes.

dhoppe:
@seanmalloy Thank you very much for your contribution.

PR #141, opened 2017-04-19
Fix duplicate declaration error

sparrovv:
Hi, I was getting:
```
Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: User[foobar] is already declared..
```
When I wanted to use already existing user, and passed it to:

```
  class { 'kafka':
    user         => 'foobar',
    group        => 'foobar',
  }
```

This should address that problem.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -37,6 +37,12 @@
 # [*package_ensure*]
 # Package version (or 'present', 'absent', 'latest'), when installing kafka from a package.
 #
+# [*manage_group*]
+# creates $group if set to true
+#
+# [*manage_user*]
+# creates $user if set to true
+#
 # [*group_id*]
 # Create kafka group with this ID
 #
@@ -67,6 +73,8 @@ class kafka (
   $package_dir    = $kafka::params::package_dir,
   $package_name   = $kafka::params::package_name,
   $package_ensure = $kafka::params::package_ensure,
+  $manage_group   = $kafka::params::manage_group,
+  $manage_user    = $kafka::params::manage_user,
   $group_id       = $kafka::params::group_id,
   $user_id        = $kafka::params::user_id,
   $user           = $kafka::params::user,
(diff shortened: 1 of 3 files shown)


juniorsysadmin:
@sparrovv Perhaps consider using ```$manage_user``` and ```$manage_group``` parameters (defaulting to ```true```) instead, as ```defined()``` is viewed as a bit of a hack

sparrovv:
@juniorsysadmin good tip, thanks!

manifests/init.pp
```diff
@@ -101,16 +109,20 @@
...
-  user { $user:
-    ensure  => present,
-    shell   => '/bin/bash',
-    require => Group[$group],
-    uid     => $user_id,
+  if $manage_user {
+    user { $user:
+      ensure  => present,
+      shell   => '/bin/bash',
+      require => Group[$group],
```
juniorsysadmin: Not to do with this PR but I can't think of a good reason for the ```require```

juniorsysadmin:
``kafka``` is hardcoded in the systemd and SysV init templates in this module as well. It would be terrific if they were not....

sparrovv:
> kafka is hardcoded as the user in the systemd and SysV init templates in this module as well. It would be terrific if they were not....

Thanks for letting me know - I'll be off for more than 3 weeks now and won't be able to address it.
We can wait until I'm back or merge this PR as it is.

manifests/init.pp
```diff
@@ -67,6 +73,8 @@
   $package_dir    = $kafka::params::package_dir,
   $package_name   = $kafka::params::package_name,
   $package_ensure = $kafka::params::package_ensure,
+  $manage_group   = $kafka::params::manage_group,
```
bastelfreak: can you please add tests for the new parameters  + resources?

LionelCons:
This has been fixed recently. Please check that `master` works fine and close this PR.

sparrovv:
> Please check that master works fine and close this PR.

Will check it today

Issue #113, opened 2016-11-01
Kafka 0.10.0.1 fails to start, inter.broker.protocol.version is set 0.8.2.2


atrepca:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: `3.7.2`
- Ruby: `2.3.1p112`
- Distribution: `Ubuntu 16.04.1 LTS`
- Module version: `2.1.0`

### How to reproduce (e.g Puppet code you use)
```
class { '::kafka::broker':
  version => '0.10.0.1',
}
```

### What are you seeing
Kafka fails to start with:
```
[2016-11-01 14:39:17,668] FATAL  (kafka.Kafka$)
java.lang.IllegalArgumentException: requirement failed: log.message.format.version 0.10.0-IV1 cannot be used when inter.broker.protocol.version is set to 0.8.2.2
```

### What behaviour did you expect instead
Kafka process should start successfully.

### Output log
Pasted above.

### Any additional information you'd like to impart
This is caused by `inter.broker.protocol.version` being set to `0.8.2.2` in params.pp on [this](https://github.com/voxpupuli/puppet-kafka/blob/ec6d1a421d7ee1ff3b6864178ce1ef5583e03e92/manifests/params.pp#L124) line. Can go around it by setting the desired version in the `::kafka::broker` class, but a better alternative is to set `'inter.broker.protocol.version' => $version,` in params.pp.

cptcanuck:
You can declare the protocol version you want when you instantiate the broker:

`
        ### default is to user inter.broker.protocol.version 8.2.2 which breaks.  Use the same version we install
        class { 'kafka::broker':
                config => { 'broker.id' => '0', 'zookeeper.connect' => 'localhost:2181', 'inter.broker.protocol.version' => '0.10.1.0' }
        }
`

atrepca:
Yup, I mentioned that too in 'additional information'. But does it make any sense to have it hardcoded to `0.8.2.2` when it can be a param, and thus one less thing to set in `kafka::broker`? Right now you have to set the version in 2 places:
```
  class { '::kafka::broker':
    version => $version,
    config  => {
      'broker.id'                     => $broker_id,
      'zookeeper.connect'             => "my.zk.server.com:2181",
      'inter.broker.protocol.version' => $version,
    }
```
[...]

PR #142, opened 2017-04-20
Add log.message.format.version to broker params

winks:
Patch against 3.0.0/master

When upgrading Kafka from `0.8` to `0.10`, the docs at
https://kafka.apache.org/documentation/#upgrade tell you to go gradually by upping `inter.broker.protocol.version` and `log.message.format.version`.

Not really related to #113 - but it also mentions rolling upgrades.

There's this Kafka issue that is kind of related https://issues.apache.org/jira/browse/KAFKA-3827 to this topic, but not to this PR.

**Caveat**: I only tested this with `0.8.2.2` and `0.10.2.0` (both work with this new parameter set), but not `0.9`.

`0.8.2.2` doesn't *know* the config values, so there are these warnings, but I think I didn't make it worse, as the `3.0.0` module already sets `inter.broker.protocol.version` and Kafka happily ignores it.
  * `WARN Property inter.broker.protocol.version is not valid (kafka.utils.VerifiableProperties)`
  * `WARN Property log.message.format.version is not valid (kafka.utils.VerifiableProperties)`

One thing I am not sure about is what the *default value* (if one doesn't configure it) is, per version. But as the current `log.message.format.version` is also `0.8.2.2` I think this is safe.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -127,6 +127,7 @@ class kafka::params {
     'group.max.session.timeout.ms'                  => '30000',
     'group.min.session.timeout.ms'                  => '6000',
     'inter.broker.protocol.version'                 => '0.8.2.2',
+    'log.message.format.version'                    => '0.8.2.2',
     'log.cleaner.backoff.ms'                        => '15000',
     'log.cleaner.dedupe.buffer.size'                => '134217728',
     'log.cleaner.delete.retention.ms'               => '86400000',


bastelfreak:
Hi @winks, thanks for the PR! @LionelCons what do you think about this one?

LionelCons:
This PR changes the default value and is fine by me.

Upgrading Kafka is indeed tricky and Puppet cannot do everything by itself. It needs a scheduler for the rolling restarts.

PR #148, opened 2017-05-22
add support to kafka::broker logs dir

jolivares:
Allow to configure Kafka app logs dir.
https://github.com/apache/kafka/blob/trunk/bin/kafka-run-class.sh:
```
# Log directory to use
if [ "x$LOG_DIR" = "x" ]; then
  LOG_DIR="$base_dir/logs"
fi
```
Despite being able to set the log4j properties file for Kafka logging, garbage collector files don't use log4j: `-Xloggc:$LOG_DIR/$GC_LOG_FILE_NAME`

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -68,6 +68,7 @@ class kafka::broker (
   $group_id                   = $kafka::params::group_id,
   $user_id                    = $kafka::params::user_id,
   $config_dir                 = $kafka::params::config_dir,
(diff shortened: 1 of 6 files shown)


jolivares:
Oops I did not notice there's a `$log_dir ` param already defined. It would be just a matter of use it

bastelfreak:
Thanks @jolivares for the PR!

PR #149, opened 2017-05-24
enable possibility of system users for kafka

Wayneoween:
This enables the usage of system users via puppet, using system user IDs.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -56,7 +56,9 @@ class kafka (
   $package_name   = $kafka::params::package_name,
   $package_ensure = $kafka::params::package_ensure,
   $group_id       = $kafka::params::group_id,
+  $system_group   = $kafka::params::system_group,
   $user_id        = $kafka::params::user_id,
+  $system_user    = $kafka::params::system_user,
 ) inherits kafka::params {
 
   validate_re($::osfamily, 'RedHat|Debian\b', "${::operatingsystem} not supported")
@@ -88,6 +90,7 @@ class kafka (
   group { 'kafka':
     ensure => present,
     gid    => $group_id,
+    system => $system_group,
   }
 
   user { 'kafka':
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -56,7 +56,9 @@
   $package_name   = $kafka::params::package_name,
   $package_ensure = $kafka::params::package_ensure,
   $group_id       = $kafka::params::group_id,
+  $system_group   = $kafka::params::system_group,
   $user_id        = $kafka::params::user_id,
+  $system_user    = $kafka::params::system_user,
```
bastelfreak: Can you add puppet4 datatypes for those two new params (and for the existing ones as well if you like)

Wayneoween: Yes I could! And I would, if I would be at a point where I can make the switch to puppet4 :sweat:
That's the reason I made this PR to the `puppet3` branch. :sweat_smile:

manifests/params.pp
```diff
@@ -17,7 +17,9 @@
   $package_name   = undef
   $package_ensure = 'present'
   $group_id       = undef
+  $system_group   = true
   $user_id        = undef
+  $system_user    = true
```
Wayneoween: Since I don't want the old behavior to change, I am going to change these new `system_`-variables to `false`.

bastelfreak:
Hi @Wayneoween. We don't accept new features anymore on the puppet3 branch. This is really really legacy and should not be used anymore. Feel free to join our IRC channel #voxpupuli on freenode if you've any questions. We're available 24/7 to help you, also with a pupet3->puppet4 migration.

LionelCons:
Note that you could also create the user and group yourself and tell the Kafka module not to manage these using the newly added `manage_user` and `manage_group` options.

Wayneoween:
Since there are no pull requests accepted to the puppet3 branch I will close this one.

With #197 I created one for puppet4, since in the meantime we have done the upgrade.

PR #153, opened 2017-06-11
replace validate_* with datatypes

bastelfreak:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -47,7 +47,7 @@ class kafka::params {
   -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.port=9993'
   $consumer_log4j_opts = $broker_log4j_opts
 
-  $limit_nofile = '65536'
+  $limit_nofile = 65536
 
   $service_restart = true
 
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,7 +12,7 @@ class kafka::params {
   # this is all only tested on Debian and RedHat
   # params gets included everywhere so we can do the validation here
   unless $facts['os']['family'] =~ /(RedHat|Debian)/ {
-    fail("${facts['os']['family']} is not supported")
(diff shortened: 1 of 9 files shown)


manifests/params.pp
```diff
@@ -8,6 +8,12 @@
 # It sets variables according to platform
 #
 class kafka::params {
+
+  # this is all only tested on Debian and RedHat
+  # params gets included everywhere so we can do the validation here
+  unless $facts['os']['family'] =~ /(RedHat|Debian)/ {
+    fail("${facts['os']['family']} is not supported")
+  }
```
hunner: We found that blocking other OSs caused issues because sometimes it WOULD work, just wasn't tested. Or it would be close so someone would take it the rest of the way. Is this going to be a standard?

alexjfisher: @bastelfreak Maybe we should change this to a warning?

types/url.pp
```diff
@@ -0,0 +1 @@
+type Kafka::Url = Pattern[/^(https?:\/\/)([\da-z\.-]+)\.([a-z\.]{2,6})(:[\d]{2,5})?(\/[\w \.-]*)*\/?$/]
```
igalic: doesn't stdlib have it's own Url type?

metadata.json
```diff
@@ -60,7 +60,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.6.1 < 5.0.0"
+      "version_requirement": ">= 4.7.0 < 5.0.0"
```
igalic: time to bump this 6.0.0 ;)

manifests/params.pp
bastelfreak: yeah

metadata.json
bastelfreak: IMO we should bump after we test against 5.X. We already discovered some problems on our zabbix module.

Issue #143, opened 2017-05-10
Package-based installation creates useless directories


LionelCons:
When specifying a `$package_name`, `init.pp` still creates the useless directories `$package_dir`, and `$install_directory` as well as the `/opt/kafka` symlink.

All of these definitions are useless in an rpm based installation and should be moved inside the `if $package_name == undef` block.

PR #167, opened 2017-07-04
create $package_dir and $install_directory ony when needed

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/143.


manifests/init.pp
```diff
@@ -176,10 +177,13 @@
       ],
       before          => File[$config_dir],
     }
+
```
bastelfreak: did you do the newlines on purpose?

LionelCons: Yes, symmetry and to get some breathing space inside the giant `if`.

Issue #144, opened 2017-05-10
Arbitrary shell variables cannot be set


LionelCons:
With a `systemd` based OS, some environment variables can be set via Puppet parameters. They are:
`KAFKA_HEAP_OPTS`, `KAFKA_LOG4J_OPTS`, `KAFKA_JMX_OPTS` and `KAFKA_OPTS`.

However, one may need to set other variables used in `kafka-run-class.sh` like `LOG_DIR` or `JAVA`.

`unit.erb` could be extended but this would be quite messy since there are many variables that could be set.

Another approach would be to use `EnvironmentFile` instead of `Environment` in `unit.erb` and allow the user to specify the content of this "environment file".

LionelCons:
Another approach would be to pass the environment variables to be set in a hash:
```
class { 'kafka::broker':
    systemd_env => {
        'LOG_DIR' => '/var/log/kafka',
        'JMX_PORT' => '9998',
    },
}
```

skpy:
> unit.erb could be extended but this would be quite messy since there are many variables that could be set.

An array could be used to permit arbitrary insertion of an unknown number of environment variables.

We just ran into this same problem: there is no way to set `LOG_DIR` in the kafka.service unit file, so logs end up in `/opt/kafka/bin/../logs`.  We're using the archive installation, so `/opt/kafka` is a symlink.  During an upgrade, the target of that symlink will change and the logs will be lost.

skpy:
Although, since we're just now dipping our toes into the Kafka waters, it doesn't look like these logs are anything other than informational (logs in the truest sense).  The `kafka-run-class.sh` creates the `LOG_DIR` directory if it doesn't exist, so starting a new version of Kafka should have that new log directory created automatically.

PR #170, opened 2017-07-07
added an $env parameter to control the environment passed to Kafka

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/144.

As a side effect, it also simplifies the templates (`init.erb` and `unit.erb`).


igalic:
I'm surprised this doesn't need a spec update

LionelCons:
The spec does not need to be updated because this change is 100% backward compatible.

Issue #176, opened 2017-07-24
mirror/service.pp should not inherit from ::kafka::params


LionelCons:
`mirror/service.pp` does not use `kafka::params` so it should not inherit from it.

Note that `mirror.pp` does use `kafka::params` and this is fine.

PR #177, opened 2017-07-24
kafka::mirror::service does not inherit from ::kafka::params anymore

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/176.


LionelCons:
It seems that `$limit_nofile` is not used consistently... I will fix this too...

bastelfreak:
Thanks for the fix! As always: Let me know when you want a new release.

Issue #178, opened 2017-07-24
include statements do not need quotes


LionelCons:
The `include` statements we use do not use quotes except in one place:
```
$ grep -r include *
manifests/broker/service.pp:      include ::systemd
manifests/consumer/service.pp:    include ::systemd
manifests/mirror/service.pp:    include ::systemd
manifests/init.pp:    include '::archive'
```

We should be consistent and never use quotes.

PR #179, opened 2017-07-24
include statements do not need quotes

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/178.


bastelfreak:
Thanks!

Issue #183, opened 2017-07-24
The ZooKeeper service should not be required by default


LionelCons:
Most Kafka nodes do not run ZooKeeper. Also, in some installations, the ZooKeeper ensemble runs on a completely different set of nodes.

So the default value for `service_requires_zookeeper` should be `false`.

PR #184, opened 2017-07-24
The ZooKeeper service is not required by default anymore

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/183.


bastelfreak:
Thanks!

Issue #173, opened 2017-07-10
Module user/group management should be optional


cptcanuck:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: NA
- Ruby: NA
- Distribution: NA
- Module version: 2/3

## How to reproduce (e.g Puppet code you use)
Use the module as it's written

## What are you seeing
The module assumes that it should be creating the kafka user and group - though it allows you to set the userid and groupid. 

## What behaviour did you expect instead
we like to manage the users and groups on the system ourselves through a separate process, which conflicts with the module trying to create the user.  We would like to be able to choose whether the module manages the user/group or not.  We'd like manage_user and manage_group options.

PR #174, opened 2017-07-10
added logic to check if user and group should be managed by the modul…

cptcanuck:
…e or externally.  https://github.com/voxpupuli/puppet-kafka/issues/173

PR #185, opened 2017-07-24
Added manage_user and manage_group parameters

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/173.

This is derived from https://github.com/voxpupuli/puppet-kafka/pull/174 that it supersedes.


bastelfreak:
Thanks!

Issue #186, opened 2017-07-24
The consumer class does not call its config subclass


LionelCons:
Just compare `manifests/producer.pp`:
```
  class { '::kafka::producer::install': }
  -> class { '::kafka::producer::config': }
  -> class { '::kafka::producer::service': }
  -> Class['kafka::producer']
```
with `manifests/consumer.pp`:
```
  class { '::kafka::consumer::install': }
  -> class { '::kafka::consumer::service': }
  -> Class['kafka::consumer']
```

LionelCons:
Hmmm... this is weirder than it looks. The consumer class does _not_ have a config subclass (like producer for instance). Instead, it has a defined type!

LionelCons:
Except `kafka::broker::topic` (which is really odd and does not relate to the broker class in any way), `kafka::consumer::config` is the only defined type used.

This makes little sense since `kafka::consumer` is a class, just like `kafka::producer::config`.

So I would recommend to change `kafka::consumer::config` from a defined type to a class.

We would have the restriction that it cannot be instantiated multiple times but we already have this restriction with all the over classes, in particular the producer class. I don't see the point in allowing multiple consumers but not multiple producers...

PR #187, opened 2017-07-24
kafka::consumer now calls kafka::consumer::config

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/186.


LionelCons:
I have changed `kafka::consumer::config` from a defined type to a class.

Note that `README.md` documented it as _both_ a class and a defined type! I guess that it started as a class (like the other "config" subclasses) and then was partially turned into a defined type.

The "config" subclasses are now all consistent. This PR is ready to be merged.

bastelfreak:
Thanks!

PR #194, opened 2017-07-28
new consumer for mirrormaker requires that there be no equal sign after whitelist or blacklist

jacobmw:
new consumer for mirrormaker requires that there be no equal sign after whitelist or blacklist


jacobmw:
Need to squash commits

PR #195, opened 2017-07-28
fix mirror maker with new consumer

jacobmw:
When using the new consumer with Mirror Maker you can not have an equals sign after either whitelist or blacklist.  When leaving the equals sign the consumer will simply ignore the paramater.


bastelfreak:
@jacobmw thanks for the PR. Could you please rebase it?

PR #196, opened 2017-07-29
MirrorMaker service needed path to .properties files

jacobmw:
Commit: 6b509b4d1373fd5c1196d6c1f350fb71f248169e changed the behaviour of the service scripts for mirror maker in such a way that they were including the config hash instead of pointing to the file containing the config.


bastelfreak:
Thanks @jacobmw !

PR #197, opened 2017-08-18
enable possibility of system users for kafka

Wayneoween:
Since we have finally gotten to puppet4 I can make another try and hope this time its correct.

Its the same as #149 but for puppet4.


bastelfreak:
Thanks!

PR #201, opened 2017-08-28
Update README.md

confiq:
adding limitation of puppet v4

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ This module is tested on the following platforms:
 * Ubuntu 12.04
 * Ubuntu 14.04
 
-It is tested with the OSS version of Puppet only.
+It is tested with the OSS version of 4< Puppet  only.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -327,7 +327,7 @@ This module is tested on the following platforms:
 * Ubuntu 12.04
 * Ubuntu 14.04
 
-It is tested with the OSS version of Puppet only.
+It is tested with the OSS version of 4< Puppet  only.
```
wyardley: Shouldn't this be "of Puppet >= 4" or similar? It looks like you may have an extra space in there as well?

Since the metadata says ">= 4.7.0", maybe the readme should match that?

confiq: did that!
I agree that it should be >=4.7 as docs say...

ekohl:
Thanks!

PR #205, opened 2017-10-04
Higher allowed version for module java and lower for zookeeper

hp197:
Tested with java module 2.1.0 and fine.
Downed zookeeper module from < 2.0 to < 1.0 to counter for sudden surprises (v0.7.5 is most current atm).

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,11 +10,11 @@
   "dependencies": [
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.0.0 <= 2.0.0"
+      "version_requirement": ">= 1.0.0 < 3.0.0"
     },
     {
       "name": "puppetlabs/java",
-      "version_requirement": ">= 1.4.2 <= 2.1.0"
+      "version_requirement": ">= 1.4.2 < 3.0.0"
     },
     {
       "name": "puppetlabs/stdlib",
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -10,19 +10,19 @@
   "dependencies": [
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.0.0 < 2.0.0"
+      "version_requirement": ">= 1.0.0 <= 2.0.0"
```
bastelfreak: Can you use '< 3.0.0'?

metadata.json
```diff
@@ -10,19 +10,19 @@
...
     },
     {
       "name": "puppetlabs/java",
-      "version_requirement": ">= 1.4.2 < 2.0.0"
+      "version_requirement": ">= 1.4.2 <= 2.1.0"
```
bastelfreak: Can you use '< 3.0.0'?

bastelfreak:
Hi @hp197, thanks for the PR! I added two inline comments.

hp197:
You put a lot of trust in the maintainers of the other modules and on the other side you want to keep maintainability yourself... Cant blame you for putting it on major revisions.

bastelfreak:
Thanks @hp197. We noticed in the past that other modules evolve very fast. We constantly need to adjust the minor versions, otherwise the puppet-module-tool isn't able to resolve the dependencies.

PR #210, opened 2017-10-17
release 4.0.0

bastelfreak:


bastelfreak:
This release has no breaking changes. The last one had some, but we only did a minor release, this should now be fixed by doing a major release. (IMO)

dhollinger:
If there were breaking changes in the last release, but a major version wasn't pushed, then this one should definitely be a maj release.

Unless there are objections, I'll approve and merge this

Issue #208, opened 2017-10-17
Kafka default values should not be used


LionelCons:
Currently, many default values for Kafka parameters are defined in `manifests/params.pp`. See for instance `$broker_config_defaults` or `$consumer_config_defaults`.

These values are merged with the user supplied values and then used to create the Kafka properties files like `server.properties`.

These values are Kafka version specific and tedious to keep up-to-date. For instance, some newer parameters like `transaction.max.timeout.ms` are missing. At the same time, these default values are useless because, by definition, default values do not need to be set. So removing all the default values should have no impact on the final configuration that Kafka will be using.

So I would recommend to stop using these `$*_config_defaults` variables or at least to remove from them all the values which are indeed Kafka defaults.

bastelfreak:
Hi @LionelCons, thanks for the explanation. Does kafka need the default values in order to start, or are they built in and we could simply drop them?

LionelCons:
AFAIK, only `zookeeper.connect` is really mandatory, all the other parameters have defaults.

However, some parameters like `broker.id` or `log.dir` (or `log.dirs`) should really be set because the defaults are very often not suitable.

LionelCons:
I have checked `manifests/params.pp` and http://kafka.apache.org/documentation.html#brokerconfigs.

All the values in `manifests/params.pp` are the documented default values that Kafka will use when not set. The only exceptions are:
 - `unclean.leader.election.enable`: `params.pp`=true, web=false (the default changed in 0.11.0.0)
 - `inter.broker.protocol.version` and `log.message.format.version`: `params.pp`='0.8.2.2' (the default is version specific)

So IMHO we should completely get rid of `$broker_config_defaults`.

As documented (http://kafka.apache.org/documentation.html#upgrade), Kafka admins should anyway control `inter.broker.protocol.version` and `log.message.format.version` when performing a cluster upgrade.

LionelCons:
I have checked `manifests/params.pp` and http://kafka.apache.org/documentation.html#consumerconfigs.

`$consumer_config_defaults` corresponds to the old consumer API that is now deprecated. The newer `$consumer_config_defaults` (commented out in our code with "Disabled because the new consumer is still beta") corresponds to the new consumer API which is now production.

Here again, we should IMHO completely get rid of `$consumer_config_defaults`.

LionelCons:
I have checked `manifests/params.pp` and http://kafka.apache.org/documentation.html#producerconfigs.

[...]

PR #213, opened 2017-10-19
remove the unneeded broker, consumer and producer default parameters

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/208.

Note that these class parameters were not documented...


bastelfreak:
Thanks @LionelCons !

Issue #218, opened 2017-10-23
Kafka default script values should not be used


LionelCons:
(this issue is similar to https://github.com/voxpupuli/puppet-kafka/issues/208)

Currently, `params.pp` contains two variables with default values to be passed to scripts:
- `$consumer_service_defaults` for `kafka-console-consumer.sh`
- `$producer_service_defaults` for `kafka-console-producer.sh`

This should not be needed. Default values should be removed and hopefully these two variables will disappear.

LionelCons:
`$consumer_service_defaults` contains defaults set in https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/tools/ConsoleConsumer.scala#L254.

`$producer_service_defaults` contains defaults set in https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/tools/ConsoleProducer.scala#L136.

PR #219, opened 2017-10-23
remove the unneeded service default parameters

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/218.


manifests/producer/service.pp
```diff
@@ -23,15 +23,14 @@
...
-  $service_defaults                          = $kafka::producer::service_defaults,
 ) {
 
   if $caller_module_name != $module_name {
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
   if $service_install {
-    $producer_service_config = deep_merge($service_defaults, $service_config)
+    $producer_service_config = $service_config
```
bastelfreak: is this assignment still needed? Can't we just work with $service_config?

LionelCons: It is currently needed because `$producer_service_config` is currently used in the ERB templates.

Cleaning this is next on my todo list ;-)

Issue #228, opened 2017-11-14
LimitNOFILE cannot be set to 'infinity'


LionelCons:
`LimitNOFILE` is controlled via `limit_nofile` which is declared as `Integer`.

So it cannot take the valid `infinity` value.

It should be declared as a `String` instead.

PR #229, opened 2017-11-14
changed limit_nofile from Integer to String

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/228.

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -128,7 +128,7 @@ class kafka::broker (
   Enum['running', 'stopped'] $service_ensure = $kafka::params::service_ensure,
   Boolean $service_restart                   = $kafka::params::service_restart,
   Boolean $service_requires_zookeeper        = $kafka::params::service_requires_zookeeper,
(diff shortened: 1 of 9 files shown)


wyardley:
Maybe a better fix would be something like `Variant[Integer, Enum['infinity', 'unlimited']]` (or whatever values are valid?

LionelCons:
@wyardley systemd's syntax also allows setting soft and hard limits at the same time with something like `LimitNOFILE=	1024:2048`. Full type validation would be too complex in this module.

alexjfisher:
Should still probably allow Integers as well as any string?

LionelCons:
@alexjfisher In this case other parameters (for instance `TimeoutStopSec` aka `timeout_stop`) should also be changed...

LionelCons:
@bastelfreak Since you approved this one, could you please merge it? Thanks!

wyardley:
I agree with @alexjfisher's suggestion about adding integer as valid too (this would also make it backwards-compatible again)... I know this is merged, but maybe someone would want to implement this as a separate PR?

Issue #237, opened 2017-11-17
The systemd script should use Wants instead of Requires


LionelCons:
`Requires` is too strict and `Wants` is the recommended way to define dependencies according to https://www.freedesktop.org/software/systemd/man/systemd.unit.html.

PR #238, opened 2017-11-17
use Wants instead of Requires

LionelCons:
This fixes https://github.com/voxpupuli/puppet-kafka/issues/237.


bastelfreak:
Thanks for the PR @LionelCons!

PR #240, opened 2017-12-04
Pin Kafka config directory mode

jnicholas1:
In some RPM packages, the Kafka config directory may not be world
readable, so Puppet should manage the permissions of both the config directory and items within. 

Let me know if you would like a specific acceptance test added for this, but I don't see file modes referenced anywhere within the existing tests.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -119,6 +119,7 @@ class kafka (
     ensure => directory,
     owner  => 'root',
     group  => 'root',
+    mode   => '0755',
   }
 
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
This should be an Optional parameter, not fixed to 0755.

bastelfreak:
Hi @jnicholas1, I'm going to close this PR due to inactivity. Please feel free to respond if you still want this merged. We need to discuss the raised questions from @vinzent and @juniorsysadmin .

PR #250, opened 2018-04-04
Add options for more sensible config permissions; puppet stdlib to 4.22.0

### LINKED ISSUE #249: Config files should not be world readable
The issues #146 and #165 already went in the right direction with making permissions more restrictive.
The default is still that config files are owned by root:root with permissions of 0644 so that the non-privileged kafka user can read the configs.
Given that keystore credentials are also specified in `server.properties` the 0644 permissions are rather broad, as anyone with access to the machine can then read the keystore credentials.
As such I'd propose being able to configure the permissions of the config files and changing the group to the configurable group parameter.
Then it's possible to have 0640 permissions with the kafka group being able to read any configs while being unable to modify anything and no other unprivileged users being able to access configs.

langesven:
This adds a `config_mode` option to be able to configure the permissions on config files. The default is set to the current default, that is, world readable config files.
Specifying the `config_mode` parameter enables one to restrict permissions to disallow `other` from reading the config files.

Fixes #249

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -55,6 +55,9 @@
 # [*manage_group*]
 # Create the kafka group if it's not already present.
 #
+# [*config_mode*]
+# The permissions for the config files.
+#
 # [*config_dir*]
 # The directory to create the kafka config files to.
 #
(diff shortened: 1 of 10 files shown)


langesven:
Didn't know about the Filemode type, that's pretty nice :)
I'm kinda at a loss why the pipelines are failing though. I don't quite see the difference to other params that seem to be magically available. We're using roughly this code in production on a local fork and it works, so I guess it's something with the tests specifically, any pointers on that?

langesven:
Missed the direct declarations in the mirror area, should be all good now.

bastelfreak:
Thanks for the PR @langesven !

PR #257, opened 2018-05-29
Some updates to be able to use an internal http mirror with proxy to download the .tar.gz archive

idrissneumann:
Adding proxy settings in params and a configurable subpath to be able to download the kafka tgz archive from an internal http mirror url (we've got this constraint on our project).

Build and tests results: https://travis-ci.org/idrissneumann/puppet-kafka/builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,26 @@ The directory to install kafka to.
 
 The url where the kafka is downloaded from.
 
+##### `mirror_subpath`
+
+The url subpath where the kafka is downloaded from (default value: `/kafka/{version}`).
+
+##### `proxy_port`
+
+The proxy port where the kafka is downloaded from.
+
+##### `proxy_host`
+
+The proxy host where the kafka is downloaded from.
+
+##### `proxy_server`
+
+The proxy server where the kafka is downloaded from (to use instead of `proxy_port` and `proxy_host` if you need to set an url as proxy for example).
+
+##### `proxy_type`
+
+The proxy type where the kafka is downloaded from (`http` for example).
+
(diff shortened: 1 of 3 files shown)


bastelfreak:
THanks for the PR @idrissneumann !

bjoernhaeuser:
I think this change makes it umpossible to download a different version then the default, the url is like this in the end `http://mirrors.ukfast.co.uk/sites/ftp.apache.org/kafka/0.11.0.1/kafka_2.11-1.1.0.tgz`, mind the different versions, `0.11.0.1` and `1.1.0`.

The full error:

```
Error: Execution of '/bin/curl http://mirrors.ukfast.co.uk/sites/ftp.apache.org/kafka/0.11.0.1/kafka_2.11-1.1.0.tgz -o /tmp/kafka_2.11-1.1.0.tgz_20180704-26913-1kntvi8 -fsSL --max-redirs 5' returned 7: curl: (7) Failed connect to mirrors.ukfast.co.uk:80; Connection refused
```

So this will still not work with a working mirror though :)

bastelfreak:
@bjoernhaeuser ah damn. Can you provide a patch for this or an acceptance test that fails? @idrissneumann can you look into this?

idrissneumann:
@bastelfreak I'll get a look

idrissneumann:
@bjoernhaeuser (cc @bastelfreak ). I think you can use `mirror_subpath` to override the default value `/kafka/{version}` in this case.

bjoernhaeuser:
@idrissneumann it is already fixed :)

PR #260, opened 2018-06-26
add support for topic configurations at creation time

TheMeier:
Optionally support the `--config` option  at topic creation

### DIFF
--- a/manifests/topic.pp
+++ b/manifests/topic.pp
@@ -7,21 +7,36 @@
 # This defined type is used to manage the creation of kafka topics.
 #
 define kafka::topic(
-  $ensure             = '',
-  $zookeeper          = '',
-  $replication_factor = 1,
-  $partitions         = 1,
-  $bin_dir            = '/opt/kafka/bin',
+  String                        $ensure             = '',
+  String                        $zookeeper          = '',
+  Variant[Integer,String]       $replication_factor = 1,
+  Variant[Integer,String]       $partitions         = 1,
+  String                        $bin_dir            = '/opt/kafka/bin',
+  Optional[Hash[String,String]] $config             = undef,
 ) {
+
+  if is_string($replication_factor) {
+    deprication('kafka::topic', 'Please use Integer type, not String, for paramter replication_factor')
+  }
+  if is_string($partitions) {
+    deprication('kafka::topic', 'Please use Integer type, not String, for paramter partitions')
+  }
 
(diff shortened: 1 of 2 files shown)


manifests/topic.pp
```diff
@@ -7,21 +7,28 @@
 # This defined type is used to manage the creation of kafka topics.
 #
 define kafka::topic(
-  $ensure             = '',
-  $zookeeper          = '',
-  $replication_factor = 1,
-  $partitions         = 1,
-  $bin_dir            = '/opt/kafka/bin',
+  String                        $ensure             = '',
```
bastelfreak: datatypes \o/

manifests/topic.pp
```diff
@@ -7,21 +7,28 @@
...
+  String                        $bin_dir            = '/opt/kafka/bin',
+  Optional[Hash[String,String]] $config             = undef,
 ) {
 
   $_zookeeper          = "--zookeeper ${zookeeper}"
   $_replication_factor = "--replication-factor ${replication_factor}"
   $_partitions         = "--partitions ${partitions}"
 
+  if $config {
+    $_config = inline_template('<% @config.each do |key, value|%> --config <%= key %>=<%= value %><% end %>')
```
bastelfreak: it should be possible to achieve the same with the `join()` function, right?

spec/defines/topic_spec.rb
```diff
@@ -18,12 +18,33 @@
         {
           'ensure'             => 'present',
           'zookeeper'          => 'localhost:2181',
-          'replication_factor' => '1',
```
bastelfreak: hrm do we need to assume that people used quoted integers? If so, the new datatype would be a breaking change.

TheMeier: How about a Variant[String, Integer]) and throw a deprecation warning on string-type?

manifests/topic.pp
TheMeier: Hm ... yes. Will change it

bastelfreak:
Hey @TheMeier, thanks for the PR. Please take a look at the inline comments I made. Also check the used email address in the commit, it isn't associated with your github account.

spec/defines/topic_spec.rb
bastelfreak: sounds good!

TheMeier:
@bastelfreak all done :)

PR #262, opened 2018-07-05
Add test for different kafka version

bjoernhaeuser:

Add a acceptance tests which should work with a different kafka version.

See comment in #257

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -21,7 +21,7 @@ class kafka::params {
   $bin_dir        = '/opt/kafka/bin'
   $log_dir        = '/var/log/kafka'
   $mirror_url     = 'https://www.apache.org/dyn/closer.lua?action=download&filename='
-  $mirror_subpath = "/kafka/${version}"
+  $mirror_subpath = "kafka/${version}"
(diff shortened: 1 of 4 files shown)


bjoernhaeuser:
@bastelfreak The test was already there, just updated it to a current kafka versiojn

manifests/params.pp
```diff
@@ -20,8 +20,8 @@
   $config_dir     = '/opt/kafka/config'
   $bin_dir        = '/opt/kafka/bin'
   $log_dir        = '/var/log/kafka'
-  $mirror_url     = 'https://www.apache.org/dyn/closer.cgi?path='
-  $mirror_subpath = "/kafka/${version}"
+  $mirror_url     = 'https://www.apache.org/dyn/closer.lua?action=download&filename='
```
bastelfreak: this looks like a bugfix that should be made in a separate pull request

bjoernhaeuser: Yeah, well, in the end you are right, but without it does not work..

bjoernhaeuser:
@bastelfreak Changed!

Issue #226, opened 2017-11-14
Systemd limits should be configurable


LionelCons:
Currently, `LimitNOFILE` is always set but its value can be controlled via the `limit_nofile` parameter. It should be possible not to set `LimitNOFILE` in the `systemd` unit file and use the system defaults.

Similarly, it should be possible to control `LimitCORE` which is currently always set to `infinity`.

PR #278, opened 2018-10-05
allow puppet/archive 3.x and puppetlabs/java 3.x

TheMeier:
update java dependency

PR #275, opened 2018-08-13
modulesync 2.11.0

bastelfreak:
modulesync 1.9.6

PR #281, opened 2019-02-08
Match params of init script with the ones of systemd unit

tooooots:

Adds the same ulimit settings in the init script as the ones that are already present in systemd unit file.

Improve #226

### DIFF
--- a/templates/init.erb
+++ b/templates/init.erb
@@ -50,9 +50,17 @@ if [ -f /etc/default/kafka ]; then
 fi
 
 start() {
-  ulimit -n 65536
+
+  <% if @limit_nofile -%>
+  ulimit -n <%= @limit_nofile %>
+  <% end -%>
+
+  <% if @limit_core -%>
+  ulimit -c <%= @limit_core %>
+  <% end -%>
+
   ulimit -s 10240
-  ulimit -c unlimited
+
   if [ -f "$PID_FILE" ]; then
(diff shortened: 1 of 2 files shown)


templates/init.erb
```diff
@@ -50,9 +50,17 @@ if [ -f /etc/default/kafka ]; then
 fi
 
 start() {
-  ulimit -n 65536
+
+  <% unless @limit_nofile.nil? -%>
```
bastelfreak: Can't we use this?
```suggestion
  <% if @limit_nofile -%>
```

bastelfreak:
I'm afraid we somehow need to get https://github.com/voxpupuli/puppet-kafka/pull/278 and https://github.com/voxpupuli/puppet-kafka/pull/275 merged before we can proceed here :(

bastelfreak:
Thanks @tooooots. Can you also add a unit test for the change?

tooooots:
thanks for the quick feedbacks @bastelfreak

bastelfreak:
Can you have a look at the failing travis jobs?

dhoppe:
@tooooots Could you please do a rebase against the ```master``` branch? Since ```modulesync``` has been updated, the ```.travis.yml``` is out of date. For example it will do the RSpec tests against Ruby 2.1.9, which seems to break Travis CI.

PR #291, opened 2020-03-19
Update default version to 2.4.1 with scala 2.12

alexjfisher:
The old versions are no longer downloadable from the default download location.
Updating to the latest version is a breaking change.

`pgrep -f` in init.d was changed due to a [4096 character limitation.](https://unix.stackexchange.com/questions/570819/pgrep-f-matches-on-first-4096-characters-only?noredirect=1)

Acceptance tests have been fixed up and testing is now also done on CentOS 6.
A lot of the effort was getting the zookeeper module to work.

Tests aren't necessarily that brilliant, (eg. multi-node tests would be better), but at least they pass now.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -61,7 +61,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.10.0 < 6.0.0"
+      "version_requirement": ">= 4.10.0 < 7.0.0"
     }
   ],
   "tags": [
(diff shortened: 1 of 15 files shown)


metadata.json
```diff
@@ -61,7 +61,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.10.0 < 6.0.0"
+      "version_requirement": ">= 4.10.0 < 7.0.0"
```
bastelfreak: this should be bumped to 5.5.whatever now?

alexjfisher: For the sake of a separate changelog entry, I was going to do this in a separate PR later.

Issue #306, opened 2020-05-06
Incorrect tag style in init.erb template causes minor bug in generated file


pmbuko:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: all
- Ruby: n/a
- Distribution: debian variants that still use /etc/init.d scripts (pre systemd)
- Module version: all

## How to reproduce (e.g Puppet code you use)
1. Apply `kafka::[broker,mirror,producer,consumer]` to a node
2. Inspect the INIT INFO section of the `/etc/init.d/kafka[-foo]` script
3. See the `Provides` line is missing a service name and the stripped newline causes `Required-Start` line to move up to the end of the same line

## What are you seeing
When applying the kafka service to a node -- whether it's broker, mirror, producer, or consumer -- the provides line at the top of the generated init script looks like
```
### BEGIN INIT INFO
# Provides:          # Required-Start:
```
## What behaviour did you expect instead
It should look like this
```
### BEGIN INIT INFO
# Provides:          kafka[-foo]
# Required-Start:
```
## Output log
n/a

## Any additional information you'd like to impart
Unless you have defined any `$service_requires`, the resulting init script works fine , but causes benign `insserv` errors: 
```
insserv: Script kafka is broken: incomplete LSB comment.
insserv: missing `Required-Start:' entry: please add even if empty.
insserv: script K01kafka provides facility # with comment sign, skipped!
```
If you've defined $service_requires, then they will be ignored since that line is combined with the Provides line above it.

dhoppe:
This issue has been solved by #307.

PR #307, opened 2020-05-06
Fix init script parameter bug

pmbuko:
Change tag format to correctly add service name to init script Provides line and prevent stripping line break

#306

### DIFF
--- a/spec/acceptance/producer_spec.rb
+++ b/spec/acceptance/producer_spec.rb
@@ -138,6 +138,10 @@ describe 'kafka::producer', if: (fact('operatingsystemmajrelease') == '6' && fac
         it { is_expected.to contain 'export KAFKA_LOG4J_OPTS="-Dlog4j.configuration=file:/opt/kafka/config/log4j.properties"' }
       end
 
+      describe file('/etc/init.d/kafka-producer'), if: (fact('service_provider') == 'upstart' && fact('osfamily') == 'Debian') do
+        it { is_expected.to contain %r{^# Provides:\s+kafka-producer$} }
+      end
+
       describe service('kafka-producer') do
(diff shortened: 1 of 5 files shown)


dhoppe:
The tests fail because ```Provides``` is only set for Debian and Ubuntu:
* https://github.com/voxpupuli/puppet-kafka/blob/master/templates/init.erb#L5

spec/acceptance/producer_spec.rb
```diff
@@ -130,14 +130,31 @@ class { 'kafka::producer':
         apply_manifest(pp, catch_changes: true)
       end
 
-      describe file('/etc/init.d/kafka-producer') do
+      describe file('/etc/init.d/kafka-producer'), if: (fact('operatingsystemmajrelease') =~ %r{(5|6)} && fact('osfamily') == 'RedHat') do
```
pmbuko, line 133: kafka-producer service was missing the same os-specific spec tests that existed for the other services.

bastelfreak:
thanks for the fix!

dhoppe:
I think this fix was not necessary, because it only affects Debian and Ubuntu. 
* https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/upstart.rb#L19

All Debian and Ubuntu releases, supported by this module, already use Systemd.

pmbuko:
> I think this fix was not necessary, because it only affects Debian and Ubuntu.
> 
> * https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/upstart.rb#L19
> 
> All Debian and Ubuntu releases, supported by this module, already use Systemd.

Ah, noted. I'm using an older version of the module and still have Ubuntu 14 hosts in production. :)

PR #313, opened 2020-05-22
Refactor unit tests

dhoppe:


spec/classes/broker_spec.rb
```diff
@@ -2,172 +2,140 @@
...
+    context "on #{os}" do
+      let(:facts) do
+        facts
       end
-    end
 
-    describe 'kafka::broker::config' do
-      context 'defaults' do
-        it { is_expected.to contain_file('/opt/kafka/config/server.properties') }
+      let :common_params do
```
ekohl, line 11: One pattern I'd suggest is (at the same level as `let(:facts)`:
```ruby
let(:params) do
  end
end
```

Then in a lower level you can use `let(:params) { super().merge(...) }`

dhoppe: I am not sure how this is supposed to look like. ```:common_params``` is used multiple times and required by ```shared_examples_param_validation.rb```.

ekohl: It's not a blocker and in itself this is probably already an improvement.

spec/classes/broker_spec.rb
```diff
@@ -2,172 +2,140 @@
...
-  it { is_expected.to contain_class('kafka::broker::install').that_comes_before('Class[kafka::broker::config]') }
-  it { is_expected.to contain_class('kafka::broker::config').that_comes_before('Class[kafka::broker::service]') }
-  it { is_expected.to contain_class('kafka::broker::service').that_comes_before('Class[kafka::broker]') }
-  it { is_expected.to contain_class('kafka::broker') }
-
-  context 'on Debian' do
-    describe 'kafka::broker::install' do
-      context 'defaults' do
-        it { is_expected.to contain_class('kafka') }
+  on_supported_os.each do |os, facts|
```
ekohl, line 5: It's better to use `os_facts` as a variable name so you don't have a let block and a variable overlap.

spec/classes/broker_spec.rb
```diff
@@ -2,172 +2,140 @@
...
-    end
 
-    describe 'kafka::broker::service' do
-      context 'manage_service false' do
-        let :params do
-          common_params.merge(manage_service: false)
-        end
+      case facts[:service_provider]
+      when 'systemd'
+        describe 'kafka::broker::service' do
```
ekohl, line 42: I'd do something like:
```suggestion
        describe 'kafka::broker::service on systemd', if: os_facts[:service_provider] == 'systemd' do
```

Later you can use `unless: ...` as well.

Alternatively you can merge them. For example, with manage_service false, you can expect it not to contain both and have a single test.

spec/classes/broker_spec.rb
dhoppe: I just fixed it, after I took a closer look at the docs.

spec/classes/broker_spec.rb
dhoppe: I would like to proceed without changing this.

spec/classes/broker_spec.rb
dhoppe: I just changed it and found an error related to ```limit_nofile``` and ```limit_core``` for non Systemd environments.

ekohl:
In https://github.com/ekohl/puppet-kafka/commit/12a27d62bb9cc2c77b7d0bea29487b585640583c I've refactored the tests slightly. It's a commit because that's much easier to show than some inline comments.

In this commit you see two styles
* explicit context blocks
* inline if

The benefit of explicit context blocks is in the test reporting. This may or may not be useful.

There's a third as well:
```ruby
it do
  if facts[:service_provider] == 'systemd'
    is_expected.to do_something
  else
    is_expected.not_to do_something
  end
end
```

As you can see, it's not great to read.

The benefit of only writing the assertions once is that you know you're testing for both cases. Otherwise you need to scroll up and down to compare. Short summary: only specify params once, assertions can have if conditions.

I'll leave it up to you to decide which style you prefer but at least you know the options.

As for params, this commit is untested, but https://github.com/voxpupuli/puppet-kafka/commit/32eae6e1e89739debdaf6f131dbaca0e4d280ffe should do it.

spec/classes/producer_spec.rb
```diff
@@ -2,78 +2,51 @@
...
+        end
+      end
 
-    describe 'kafka::producer::service' do
-      context 'defaults' do
-        it { is_expected.to raise_error(Puppet::Error, %r{Console Producer is not supported on systemd, because the stdin of the process cannot be redirected}) }
+      describe 'kafka::producer::service' do
+        context 'defaults' do
+          if os_facts[:service_provider] == 'systemd'
+            it { is_expected.to raise_error(Puppet::Error, %r{Console Producer is not supported on systemd, because the stdin of the process cannot be redirected}) }
```
ekohl, line 41: It was there before, but you can also use
```suggestion
            it { is_expected.to compile.and_raise_error(%r{Console Producer is not supported on systemd, because the stdin of the process cannot be redirected}) }
```

PR #316, opened 2020-06-03
Add support for Hiera

dhoppe:


data/family/Debian.yaml
```diff
@@ -0,0 +1,10 @@
+---
+kafka::user_shell: '/usr/sbin/nologin'
```
bastelfreak, line 2: I don't think that all those data duplication is perfect, but I think it can be cleaned up in the future when all/more is moved to hiera and we've a better ida of the general structure.

dhoppe: I agree. This is related to stuff like: 
* https://github.com/voxpupuli/puppet-kafka/blob/master/manifests/broker/install.pp#L10-L30

PR #320, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0


ekohl (changes requested):
Needs puppet-lint fixes.

Also looks like EL7 fails on Zookeeper (maybe EL6 too).
```
Error: Execution of '/opt/puppetlabs/puppet/bin/curl http://apache.org/dist/zookeeper/zookeeper-3.6.1/apache-zookeeper-3.6.1-bin.tar.gz -o /tmp/apache-zookeeper-3.6.1-bin.tar.gz_20201125-1171-1p90w3b -fsSLg --max-redirs 5' returned 22: curl: (22) The requested URL returned error: 404 Not Found
```

.sync.yml
```diff
@@ -6,7 +6,6 @@
     - set: debian10-64
```
ekohl, line 6: It looks like this entire file can be removed now.

manifests/params.pp
```diff
@@ -6,7 +6,7 @@
     warning("${facts['os']['family']} is not supported")
   }
 
-  $kafka_version  = '2.4.1'
+  $kafka_version  = '2.7.0'
```
ekohl, line 9: This is more than just a modulesync. That should be noted in somewhere in the changelog. The best way to do that is to open an issue for it.

metadata.json
```diff
@@ -33,14 +33,12 @@
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "6",
```
ekohl, line 36: This should also be noted in the changelog. The best way to do that is to open an issue for it.

bastelfreak: since that's coming from https://github.com/voxpupuli/puppet-kafka/pull/322 I think we don't need a separate issue?

ekohl: Only if that's merged. If it's just closed it won't show up.

manifests/params.pp
bjoernhaeuser: I reverted this change and simply using the "apache archive", which will make sure that the test will continue to run in the future.

.sync.yml
bjoernhaeuser: Removed this file in b898cf7bfe8b406b277bc313baf047a0076cfd04

metadata.json
bjoernhaeuser: Please let me know what comes out of this discussion.

manifests/params.pp
```diff
@@ -12,7 +12,7 @@
   $config_dir     = '/opt/kafka/config'
   $bin_dir        = '/opt/kafka/bin'
   $log_dir        = '/var/log/kafka'
-  $mirror_url     = 'https://www.apache.org/dyn/closer.lua?action=download&filename='
+  $mirror_url     = 'https://archive.apache.org/dist/'
```
ekohl, line 15: I think upgrading to the latest version is a better thing. I don't think archive is mirrored so it's not very nice to default to it. I don't mind updating the default version and probably would even encourage it but that does deserve a changelog entry. As I said: opening an issue is the best way to do so.

metadata.json
ekohl: If you rebase this PR on top of #322, it'll automatically be marked as merged when this PR is merged. That's a strategy we've used multiple times and works best well.

ekohl (changes requested):

manifests/params.pp
bjoernhaeuser: I disagree. This could simply make the tests fail tomorrow whenever "apache" believes it's okay to remove the old versions. This happens to zookeeper as well and 3.6.1 is not very old, but still gone.

metadata.json
bjoernhaeuser: It is rebased on #322.

manifests/params.pp
ekohl: Fair enough, but I would consider it something that must be noted in the changelog. It changes default behavior and as a user, that's something of note. Remember that modulesyncs are left out of the changelog by default and we don't want to require users to read the git log. That's why this is important to get right.

bjoernhaeuser: Okay, my suggestion:

* Merge this PR to get "green test" base again
* Update to kafka 2.7.0
* Move back to old mirror_url again
* Optionally: Use the archive in testing, not sure how this works.

I can carry out 1-3 if you agree.

ekohl: I don't understand why opening an issue that states the new behavior and linking it in this PR is too hard. It is trivial and solves the problem.

bjoernhaeuser:
Also solves/closes this #323

manifests/params.pp
bjoernhaeuser: Opened the issue and cross referenced.

ekohl: Thanks!

ekohl (changes requested):
There's also a bunch of Red Hat 6 code left in acceptance tests:
https://github.com/voxpupuli/puppet-kafka/blob/70ec701dbe4de76a376bca3023101cf228fc5832/spec/acceptance/01_zookeeper_spec.rb#L21-L24
https://github.com/voxpupuli/puppet-kafka/blob/70ec701dbe4de76a376bca3023101cf228fc5832/spec/acceptance/broker_spec.rb#L3-L12

There's a few more. Would you mind cleaning that up as well?

Other than that, :+1:

spec/acceptance/01_zookeeper_spec.rb
```diff
@@ -18,16 +18,12 @@ class { 'java' :
...
-      $zookeeper_service_provider = $facts['os']['release']['major'] ? {
-        '6' => 'redhat',
-        '7' => 'systemd',
-      }
-
       class { 'zookeeper':
         install_method      => 'archive',
         archive_version     => '3.6.1',
-        service_provider    => $zookeeper_service_provider,
+        service_provider    => 'systemd',
```
ekohl, line 24: Looks like this was already supposed to work without passing it: https://github.com/deric/puppet-zookeeper/blob/4cf47591dde9f6efbde1635005c68cdf515b26cd/manifests/params.pp#L32-L44 so doesn't this also work:
```suggestion
```

spec/acceptance/producer_spec.rb
```diff
@@ -1,162 +0,0 @@
...
-else
-  case fact('osfamily')
-  when 'RedHat', 'Suse'
-    user_shell = '/sbin/nologin'
-  when 'Debian'
-    user_shell = '/usr/sbin/nologin'
-  end
-end
-
-describe 'kafka::producer', if: (fact('operatingsystemmajrelease') == '6' && fact('osfamily') == 'RedHat') do # systemd systems not supported by kafka::producer::service
```
ekohl, line 14: Is it still true that the producer doesn't work on systemd? If you'd keep it in, would it work? If not, what (supported) code would actually call `kafka::producer`?

bjoernhaeuser: If you would be able to wire up the stdin of the producer with something it would work. Though I do not know how this could work. `kafka::producer` could only be called by a user from the module. Not sure what todo with it.

ekohl: With systemd socket activation I think that might be possible. You can connect a socket to stdin. Perhaps an issue should be opened for that?

If I look through metadata.json, I'm only unsure about SLES but otherwise all distros run systemd. That can suggest to drop the producer code altogether as well.

bjoernhaeuser: I created this one here: https://github.com/voxpupuli/puppet-kafka/issues/324

ekohl (approved):

PR #321, opened 2021-02-04
No longer accept deprecated String params

der-eismann:
Fixes the warning

```
Warning: This method is deprecated, please use match expressions with Stdlib::Compat::String instead. 
They are described at https://docs.puppet.com/puppet/latest/reference/lang_data_type.html#match-expressions. 
at ["/opt/puppetlabs/puppet/modules/kafka/manifests/topic.pp", 43]:
```

Syntax is supported from Puppet 5.5 to 7.3, so there shouldn't be any problems.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1577,7 +1577,7 @@ Default value: ''
 
 ##### `replication_factor`
 
-Data type: `Variant[Integer,String[1]]`
+Data type: `Integer`
 
 The replication factor for each partition in the topic being created. If
 not supplied, defaults to the cluster default.
@@ -1586,7 +1586,7 @@ Default value: 1
 
 ##### `partitions`
 
-Data type: `Variant[Integer,String[1]]`
+Data type: `Integer`
 
 The number of partitions for the topic being created or altered. If not
 supplied for create, defaults to the cluster default.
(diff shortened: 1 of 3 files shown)


kenyon:
Maybe it's time to just change those data types to `Integer` and don't accept strings.

Tests are failing because of a download failure:

```
Error: Execution of '/opt/puppetlabs/puppet/bin/curl https://www.apache.org/dyn/closer.lua?action=download&filename=kafka/2.4.1/kafka_2.12-2.4.1.tgz -o /tmp/kafka_2.12-2.4.1.tgz_20210204-3872-o317sh -fsSLg --max-redirs 5' returned 22: curl: (22) The requested URL returned error: 404 Not Found
```

igalic (approved):

alexjfisher:
Given how strings were deprecated almost 3 years ago in https://github.com/voxpupuli/puppet-kafka/commit/e47560a953858641a34455928258f08f45eef545 *and* that was broken and was left broken for 2 years until https://github.com/voxpupuli/puppet-kafka/commit/4508faba3ab4c8c6300a6d58961ed048c67a43f2, *and* how the next release is already going to be a major version, I also would prefer removing this code and only accepting Integers from now on.

der-eismann:
I gave it a try. Is this correct?

alexjfisher:
> I gave it a try. Is this correct?

Looks good. I had just opened https://github.com/voxpupuli/puppet-kafka/pull/325 ;)
If you can squash your commits and the build passes, I'll merged this PR instead.

alexjfisher (approved):

PR #326, opened 2021-04-16
allow latest dependencies

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,15 +10,15 @@
   "dependencies": [
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.0.0 < 5.0.0"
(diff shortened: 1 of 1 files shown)


bastelfreak:
requires a new puppetlabs-java release

martin-koerner:
You are probably already aware of this, but just to be sure: systemd v3.0.0 breaks e.g. broker::service because it still references `Exec['systemctl-daemon-reload']` which was removed in https://github.com/camptocamp/puppet-systemd/pull/171.

kenyon:
I think this needs https://github.com/deric/puppet-zookeeper/pull/171 now.

kenyon (approved):
Tests now pass!

smortex (approved):

PR #327, opened 2021-05-18
Removing SystemD reloading step

### LINKED ISSUE #324: kafka::producer class does not work with systemd
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5 + 6 
- Ruby:
- Distribution:
- Module version: 7.x

## How to reproduce (e.g Puppet code you use)

Try to use the kafka::producer class on a systemd os.

## What are you seeing

There will  not be a service installed

## What behaviour did you expect instead

A unit/service is installed and I am able to pipe message through it.

## Any additional information you'd like to impart

The console kafka::producer reads from stdin and publishes it to kafka. Systemd normally prevents that and some systemd magic would be required to make this happen. I am not sure whats the best way for this is.

As someone who is using kafka quite a bit I do not see any useful option in having a systemd unit which runs the producer, which is IMHO clearly a "test out kafka tool" or something which is only used once in a while directly from the console. 

Nevertheless the current class should either be amended to support systemd or be removed.

evgenkisel:
Removing SystemD reloading step by a special class from the camptocamp/systemd module, because the class was deprecated by the change [171](https://github.com/camptocamp/puppet-systemd/pull/171)

Fixes #324

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "camptocamp/systemd",
-      "version_requirement": ">= 0.4.0 < 3.0.0"
+      "version_requirement": ">= 0.4.0 <= 3.0.0"
     }
   ],
   "operatingsystem_support": [
--- a/metadata.json
+++ b/metadata.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "camptocamp/systemd",
-      "version_requirement": ">= 0.4.0 <= 3.0.0"
+      "version_requirement": ">= 0.4.0 < 4.0.0"
     }
   ],
   "operatingsystem_support": [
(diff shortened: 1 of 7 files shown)


metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "camptocamp/systemd",
-      "version_requirement": ">= 0.4.0 < 3.0.0"
+      "version_requirement": ">= 0.4.0 <= 3.0.0"
```
kenyon, line 29: This is usually how we do it:

```suggestion
      "version_requirement": ">= 0.4.0 < 4.0.0"
```

smortex, line 29: I think we want to change the minimum version here:
```suggestion
      "version_requirement": ">= 3.0.0 < 4.0.0"
```

evgenkisel: @kenyon thanks for the suggestion. Committed.

PR #328, opened 2021-07-20
Allow to manage Log4j configuration

### LINKED ISSUE #85: systemv init created log files are not being rotated
on a systemv platform the log files created by our init scripts are only ever written to but never rotated. Additionally, a restart will cause the existing file to be overwritten.

There are a couple of ways to solve this:
- add a logrotate entry
- use logger(1) — that way we're ~systemd compatible~
- don't redirect stdout/stderr — instead change the `log4j.rootLogger=INFO, stdout`

all of those come with advantages/drawbacks:
- the first option adds a potential dependency on a third-party module
- the second option is most simple
- the last point seems most sensible

If we go with the last, we should add a symlink between `/var/log/kafka` and `/opt/kafka/logs` — for compatibility.

### LINKED ISSUE #200: Enable customized log4j.properties files
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: current
- Ruby: current
- Distribution: current
- Module version: current

## How to reproduce (e.g Puppet code you use)
Let Kafka run for several days.

## What are you seeing
Logfiles are rotated but never deleted. Even logrotate does not work in this case.

## What behaviour did you expect instead
Possibility to configure the logfiles retention in this puppet module

## Output log
n/a

evgenkisel:
A default configuration of log4j does not allow Kafka to delete old log files which were rotated.
As result it gets issues with disk space on Kafka instances.
The changes will allow to update `log4j.properties` files to change a logic of rotation for deleting old files.
Customers can enable the function and customize a size which triggers rotating and a count of stored log files.

Fixes #200
Fixes #85


templates/log4j.properties.erb
```diff
@@ -0,0 +1,102 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
bastelfreak, line 1: could you please convert this into an epp template? We prefer this for new templates since that's the erb successor.

evgenkisel: Yes, sure. Converted

smortex (changes requested):
Can you add unit tests that demonstrate that this works as intended so that we can avoid regressions in the future?

No need to test for all values in the template, I guess that checking that the file is properly managed depending on the value of `$manage_log4j` and that MaxFileSize / MaxBackupIndex are set to the expected values for an appender would be enough.

evgenkisel:
@smortex will do.

evgenkisel:
@smortex could you please provide docs or examples how to test values of class parameters with rspec-puppet?
I'm not a big expert in Puppet Rspec.

I found how to test the file resource when `manage_log4j` is true
```
context 'with  manage_log4j => true' do
  let(:params) { {'manage_log4j' => true} }
    it { is_expected.to contain_file('/opt/kafka/config/log4j.properties') }
end
```
but  I can't find ways how to test the parameters `$log_file_size`  and `$log_file_count`
for the `$log_file_size` we should be able that it's set to `String` in format `999MB`

spec/classes/broker_spec.rb
```diff
@@ -31,6 +36,10 @@
         context 'defaults' do
           it { is_expected.to contain_file('/opt/kafka/config/server.properties') }
         end
+        context 'with  manage_log4j => true' do
+          let(:params) { {'manage_log4j' => true} }
+          it { is_expected.to contain_file('/opt/kafka/config/log4j.properties') }
```
smortex, line 41: You can test [resource parameters](https://rspec-puppet.com/documentation/classes/#test-resource-parameters) like so:

```suggestion
          it { is_expected.to contain_file('/opt/kafka/config/log4j.properties').with_content(%r{^log4j.appender.kafkaAppender.MaxFileSize=50MB$}) }
```

What you want to test is the content parameter of the file, and it's a regular String (the return value from the `epp()` call after applying the template).

You can pass non-default values in a context and make sure they are processed as expected.

evgenkisel:
@smortex  thanks a lot for your explanation! Looks like I didn't understand a main goal of the Unit tests.
So, I updated Unit tests and deleted my changes from `kafka::producer` class, because the class doesn't run a daemon of Kafka as I understand. It's just for running Kafka producer console client. 
The parameter `$manage_log4j` will be available only for the broker, consumer and mirror classes.

evgenkisel:
@smortex could you please review it? I hope we can include the changes to the new release

manifests/broker.pp
```diff
@@ -150,6 +150,9 @@
   String[1] $jmx_opts                        = $kafka::params::broker_jmx_opts,
   String[1] $log4j_opts                      = $kafka::params::broker_log4j_opts,
   String[0] $opts                            = $kafka::params::broker_opts,
+  Boolean $manage_log4j                      = $kafka::params::manage_log4j,
+  Pattern[/[0-9]+[KMG]B/] $log_file_size     = $kafka::params::log_file_size,
```
smortex, line 154: I guess `0MB` is not a valid value?

```suggestion
  Pattern[/[1-9][0-9]*[KMG]B/] $log_file_size     = $kafka::params::log_file_size,
```

smortex (commented):
Looks good, I added an in-line note for some point you may want to adjust.

Can you then squash your changes in a single commit (to avoid multiple "fix" commits which make digging in the code history harder)?

Thanks!

evgenkisel:
@smortex thank you for the suggest about changing `Pattern` for the `$log_file_size` !

all changes are squashed to [a13a5ed
](https://github.com/voxpupuli/puppet-kafka/pull/328/commits/a13a5ede0facc9648a5921a4c5b1046832c452f3)

please review it

smortex:
I still see a lot of commits :smile: 

Your commits are interleaved with other ones in the master branch (you seem to have merged master at some point), so first rebase you changes on top of master:

```
git rebase origin/master
```

You then end up with your 13 commits on top of master. Run:

```
git rebase -i origin/master
```

This will open your editor with the commits where you can reorder / edit / reword / etc.  If you just want to keep the commit message from the fist commit and squash all other commits into it, keep he first line untouched and change all other lines' "pick" to "fixup", save and exit, and you are ready to:

```
git push -f
```

evgenkisel:
@smortex  I'm sorry, but I'm not a big expert with Git :)) Now it's ok

.fixtures.yml
```diff
@@ -4,6 +4,6 @@ fixtures:
     java: 'https://github.com/puppetlabs/puppetlabs-java.git'
     stdlib: 'https://github.com/puppetlabs/puppetlabs-stdlib.git'
     zookeeper: 'https://github.com/deric/puppet-zookeeper.git'
-    systemd: 'https://github.com/camptocamp/puppet-systemd.git'
+    systemd: 'https://github.com/voxpupuli/puppet-systemd.git'
```
smortex, line 7: This is already in the `master` branch :-)

```
git fetch --all
git rebase origin/master
git push -f
```

and we should be good!

evgenkisel: I  did it but there is nothing to update
```
$ git fetch --all
Fetching origin
$ git rebase origin/master
Current branch log4j is up to date.
$ git push -f
Everything up-to-date
```

evgenkisel: Maybe  it needs to be reverted?

evgenkisel: I forgot to fetch upstream :-) 
It should be ok

smortex: No worries, thanks!

evgenkisel: thank you!

smortex (approved):
LGTM.  If nobody from voxpupuli who use this module approve and merge this, I am fine with merging this in a few days if no concerns are raised.

Thanks!

manifests/broker.pp
```diff
@@ -115,41 +115,44 @@
...
+  Boolean $daemon_start                         = $kafka::params::daemon_start,
+  Hash $env                                     = {},
+  Hash[String[1], Any] $config                  = {},
+  String[1] $heap_opts                          = $kafka::params::broker_heap_opts,
+  String[1] $jmx_opts                           = $kafka::params::broker_jmx_opts,
+  String[1] $log4j_opts                         = $kafka::params::broker_log4j_opts,
+  String[0] $opts                               = $kafka::params::broker_opts,
+  Boolean $manage_log4j                         = $kafka::params::manage_log4j,
+  Pattern[/[1-9][0-9]*[KMG]B/] $log_file_size   = $kafka::params::log_file_size,
+  Integer $log_file_count                       = $kafka::params::log_file_count
```
bastelfreak, line 155: For the integer here and in the other classes, can you enforce the minimal/maximal values?
example to allow 0 as lowest value and 10 as highest:
```suggestion
  Integer[0, 10] $log_file_count                = $kafka::params::log_file_count
```

evgenkisel: @bastelfreak Updated like `Integer[1, 50] $log_file_count`

evgenkisel:
@bastelfreak @smortex  I'm sorry, but after my last changes it started two workflows and they are failed.
As I see it's not related with my changes. Maybe some conflicts between containers on Travis CI.
could you please re-run tests?

bastelfreak (approved):

PR #331, opened 2021-07-27
Release 8.0.0

### LINKED ISSUE #330: New release
Is it posible to push new release, so we can migrate to `voxpupuli/puppet-systemd`. Since the current version prohibits from doing so because of the `Exec[systemctl-daemon-reload]` no longer available in newer versions of the `systemd` module

smortex:

Fixes #330


bastelfreak:
I will merge this and run the travis_release task locally

PR #332, opened 2021-08-05
modulesync 4.1.0

bastelfreak:
modulesync 4.1.0


bastelfreak:
merging as puppet 7 / debian 8 is expected to fail

PR #333, opened 2021-08-05
Drop EoL Puppet 5; Add Puppet 7 support

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -69,7 +69,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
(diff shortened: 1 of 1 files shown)


bastelfreak:
I would like to have this in the next major release as well

metadata.json
```diff
@@ -69,7 +69,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
+      "version_requirement": ">= 5.5.8 < 8.0.0"
```
smortex, line 72: This does not drop support for Puppet 5 as the PR title suggest :wink:

bastelfreak: fixed it

PR #334, opened 2021-08-05
Drop EoL Debian 8, Debian 9, Ubuntu 16.04 support

bastelfreak:

PR #337, opened 2022-03-11
Support kafka 3.0 topics

### LINKED ISSUE #302: [Topic] Parameter --zookeeper
> The bin/kafka-topics.sh command line tool is now able to connect directly to brokers with --bootstrap-server instead of zookeeper. The old --zookeeper option is still available for now. Please read [KIP-377](https://cwiki.apache.org/confluence/display/KAFKA/KIP-377%3A+TopicCommand+to+use+AdminClient) for more information.

der-eismann:

In kafka 3.0 the `--zookeeper` option was removed for topics. It was deprecated in v2.2.0 (March 2019) and `--bootstrap-server` has been available since then. I'm unsure if we should replace `zookeeper` with `bootstrap-server` completely or allow both options, but in that case I don't know how to require at least one of both arguments. Open for discussion here.

Fixes #302

### DIFF
--- a/manifests/topic.pp
+++ b/manifests/topic.pp
@@ -14,7 +14,10 @@
 #
 # @param zookeeper
 #   The connection string for the ZooKeeper connection in the form host:port.
-#   Multiple hosts can be given to allow fail-over.
+#   Multiple hosts can be given to allow fail-over. Kafka < 3.0.0 only!
+#
+# @param bootstrap_server
+#   The Kafka server to connect to in the form host:port. Kafka >= 2.2.0 only!
 #
 # @param replication_factor
 #   The replication factor for each partition in the topic being created. If
@@ -32,16 +35,28 @@
 #   See the Kafka documentation for full details on the topic configs.
 #
 define kafka::topic (
-  String[1] $ensure                           = '',
-  String[1] $zookeeper                        = '',
+  Optional[String[1]] $ensure                 = undef,
+  Optional[String[1]] $zookeeper              = undef,
+  Optional[String[1]] $bootstrap_server       = undef,
   Integer   $replication_factor               = 1,
   Integer   $partitions                       = 1,
   String[1] $bin_dir                          = '/opt/kafka/bin',
   Optional[Hash[String[1],String[1]]] $config = undef,
 ) {
   $_zookeeper          = "--zookeeper ${zookeeper}"
+  $_bootstrap_server   = "--bootstrap-server ${bootstrap_server}"
   $_replication_factor = "--replication-factor ${replication_factor}"
   $_partitions         = "--partitions ${partitions}"
+
+  if !$zookeeper and !$bootstrap_server {
+    fail('Either zookeeper or bootstrap_server parameter must be defined!')
+  }
+
+  if $zookeeper {
+    $_connection = $_zookeeper
+  } else {
+    $_connection = $_bootstrap_server
+  }
 
   if $config {
     $_config_array = $config.map |$key, $value| { "--config ${key}=${value}" }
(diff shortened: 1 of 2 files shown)


root-expert (changes requested):
Hello and thank you for the PR!

Take a look at the inline comments.

Regarding dropping the `zookeeper` option:
Given that ZooKeeper and bootstrap server are two different subsystems I suggest that we keep both of the options for now. Also, some versions of Kafka that can use `zookeeper` option are not yet EoL (e.g 2.4.x).

To have one of the two options defined i think it can be achieved with:

```puppet
if !$zookeeper && !$bootstrap_server {
  fail('Either zookeeper or bootstrap_server parameter must be defined!')
}
```

manifests/topic.pp
```diff
@@ -33,16 +36,24 @@
 #
 define kafka::topic (
   String[1] $ensure                           = '',
-  String[1] $zookeeper                        = '',
+  String    $zookeeper                        = '',
+  String    $bootstrap_server                 = '',
```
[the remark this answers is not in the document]
smortex: Also update default to `= undef`, instead of `= ''` which is no longer a valid value :wink:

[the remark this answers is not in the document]
root-expert: Lol, github went brrr i had proposed that haha
Thanks @smortex :smile:

der-eismann:
Thanks for the quick replies! I added all the changes, looks better now :slightly_smiling_face: 
Regarding the failing test I think #334 needs to be merged because there are no puppet 7 packages for Debian Jessie (8).

manifests/topic.pp
```diff
@@ -32,17 +35,29 @@
 #   See the Kafka documentation for full details on the topic configs.
 #
 define kafka::topic (
-  String[1] $ensure                           = '',
-  String[1] $zookeeper                        = '',
+  String[1] $ensure                           = undef,
```
root-expert, line 38: This needs to wrapped into `Optional` since it's defaulting to `undef` now
Also this could be narrowed down to an Enum. e.g:

```suggestion
Optional[Enum['present', 'absent']] $ensure   = undef
```

Although absent won't remove the topic since there's not functionality for that :smile:

manifests/topic.pp
```diff
@@ -32,17 +35,29 @@
...
+  Optional[String[1]] $zookeeper              = undef,
+  Optional[String[1]] $bootstrap_server       = undef,
```
root-expert, line 40: Allowing Strings here could a potential security risk due to the `exec` resource if I'm not mistaken.
An issue for another PR, but this could be limited/sanitized with a custom data type

manifests/topic.pp
```diff
@@ -32,17 +35,29 @@
...
   Optional[Hash[String[1],String[1]]] $config = undef,
 ) {
   $_zookeeper          = "--zookeeper ${zookeeper}"
+  $_bootstrap_server   = "--bootstrap-server ${bootstrap_server}"
   $_replication_factor = "--replication-factor ${replication_factor}"
   $_partitions         = "--partitions ${partitions}"
 
+  if !$zookeeper and !$bootstrap_server {
+    fail('Either zookeeper or bootstrap_server parameter must be defined!')
+  }
```
root-expert, line 53: Please add test case for this :smile:

root-expert (changes requested):

manifests/topic.pp
der-eismann: As you said, I think this is out of scope for this PR :slightly_smiling_face:

spec/defines/topic_spec.rb
```diff
@@ -25,6 +25,39 @@
...
+        let :params do
+          {
+            'ensure'             => 'present',
+            'replication_factor' => 1,
+            'partitions'         => 1
+          }
+        end
+
+        it {
+          is_expected.to compile.and_raise_error('Either zookeeper or bootstrap_server parameter must be defined!')
```
root-expert, line 57: Maybe wrapping this in a regex will fix the spec error

der-eismann: I did this before, but misinterpreted the linting error regarding the `%r{}`. Should be fine now!

root-expert (approved):
Please squash your commits before merging :smile:

der-eismann:
> Please squash your commits before merging smile

Done!

PR #350, opened 2023-11-23
Remove legacy top-scope syntax

smortex:


smortex:
PR titles adjusted.  Will be visible in next release.

PR #352, opened 2024-02-10
correct legacy fact ::osfamily in zookeeper spec

zilchms:
prerequisite for #348


spec/acceptance/01_zookeeper_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 describe 'zookeeper prereq' do
   zookeeper = <<-EOS
-    if $::osfamily == 'RedHat' {
+    if $facts[os][family] == 'RedHat' {
```
bastelfreak, line 7:
```suggestion
    if $facts['os']['family'] == 'RedHat' {
```

bastelfreak (approved):

PR #354, opened 2024-02-10
Add Rocky, EL 8 and 9 support

zilchms:


kenyon (commented):
I think we usually add Alma and Rocky too. Example: https://github.com/voxpupuli/puppet-autofs/pull/198/files

zilchms:
For OracleLinux and AlmaLinux the tests are failing, so I would like to add them later in a separate PR. I just dont know enough about Kafka to fix that any time soon. Ideally I would like to fix up the module first before tackling failing OS specific things

PR #357, opened 2024-02-11
Drop support for non systemd OS

zilchms:


manifests/broker/service.pp
```diff
@@ -36,31 +36,14 @@
...
-      File["/etc/systemd/system/${service_name}.service"]
-      ~> Service[$service_name]
-    } else {
-      file { "/etc/init.d/${service_name}":
-        ensure  => file,
-        mode    => '0755',
-        content => template('kafka/init.erb'),
-        before  => Service[$service_name],
-      }
+    file { "/etc/systemd/system/${service_name}.service":
```
bastelfreak, line 41: In the future we should migrate this to systemd::unit_file

manifests/producer/service.pp
```diff
@@ -37,16 +37,7 @@
...
-      fail('Console Producer is not supported on systemd, because the stdin of the process cannot be redirected')
-    } else {
-      file { "/etc/init.d/${service_name}":
-        ensure  => file,
-        mode    => '0755',
-        content => template('kafka/init.erb'),
-        before  => Service[$service_name],
-      }
-    }
+    fail('Console Producer is not supported on systemd, because the stdin of the process cannot be redirected')
```
bastelfreak, line 40: I am wondering if this is still valid

manifests/broker/config.pp
```diff
@@ -24,7 +24,6 @@
     $config_notify = undef
   }
 
-  $doctag = 'brokerconfigs'
```
bastelfreak, line 27: Is that a dead variable that isn't used anywhere?

manifests/producer/service.pp
zilchms: Thats what I have been wondering, too. But I am not too familiar with kafka to make a call there.

manifests/broker/config.pp
zilchms: Ahhh, that one I messed up I think. It is in the properties.erb but only to set the documentation link. So maybe we should keep that in?

manifests/broker/service.pp
zilchms: Yes we should. But I recon that should maybe be another PR

manifests/broker/config.pp
bastelfreak: Yeah I think we should keep it

spec/acceptance/consumer_spec.rb
```diff
@@ -140,15 +140,6 @@ class { 'kafka::consumer':
...
-      describe file('/etc/init.d/kafka-consumer'), if: (fact('service_provider') == 'upstart' && fact('osfamily') == 'Debian') do
-        it { is_expected.to be_file }
-        it { is_expected.to be_owned_by 'root' }
-        it { is_expected.to be_grouped_into 'root' }
-        it { is_expected.to contain %r{^# Provides:\s+kafka-consumer$} }
-        it { is_expected.to contain 'export KAFKA_JMX_OPTS="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.port=9993"' }
-        it { is_expected.to contain 'export KAFKA_LOG4J_OPTS="-Dlog4j.configuration=file:/opt/kafka/config/log4j.properties"' }
-      end
-
       describe file('/etc/systemd/system/kafka-consumer.service'), if: (fact('operatingsystemmajrelease') == '7' && fact('osfamily') == 'RedHat') do
```
ekohl, line 143: Would this now be true for any OS? And shouldn't this have been modified when EL 8 & EL 9 support was added?

manifests/broker/service.pp
```diff
@@ -36,31 +36,14 @@
...
-        ensure => absent,
-      }
-
-      File["/etc/systemd/system/${service_name}.service"]
-      ~> Service[$service_name]
-    } else {
-      file { "/etc/init.d/${service_name}":
-        ensure  => file,
-        mode    => '0755',
-        content => template('kafka/init.erb'),
```
jay7x, line 58: I guess this template (`kafka/init.erb`) should be deleted also then

manifests/producer/service.pp
jay7x: I'd vote for keeping it as `fail()`.. then I'd vote to drop both `kafka::consumer` and `kafka::producer` classes as those are mostly useless from my point of view. Though, I might be wrong 🤷🏻‍♂️

zilchms: I see 2 Options here:
1. Kafka producers are now supported for systemd by kafka -> this has to be patched to follow this
2. Kafka producers still dont support systemd -> the producer class class can be dropped
Both requires knowledge about kafka is dont have yet. Might be something we should solve in a different PR (maybe someone using kafka stumbles upon this module and helps out). For the time being we can keep the fail() part as we change nothing about the original logic (and therefore dont break stuff that hasnt been broken).

As for the consumer: I want to provide a PR that at least combines the subclasses together. The class in and of itself is usefull. Supporting Kafka consumers is a good feature.

jay7x: As far as I see there is nothing to do from the Kafka side. All the "magic" was implemented in the init-script completely (see `kafka::producer::input` IIRC). Theoretically, it's possible to implement something alike with the `StandartInput{,Text,Data}` unit params.

Fun enough, there is no "opposite way" for the `kafka::consumer`. That's why I cannot understand what is the use-case for the consumer class 🤔

spec/acceptance/consumer_spec.rb
zilchms: looking back at this, i looked at the wrong line. And yes, this maybe should have been modified some time ago and now it should be a check for all our supported OSes

Issue #381, opened 2024-12-17
consumer and producer classes needed?


schustersv:
Hello,
I'm currently working on some overhaul of puppet-kafka (bring puppet-systemd usage up to date, improve testing, get rid of erb templates) and have made quite some progress in updating puppet-systemd usage for the kafka::broker class.

Before putting some more effort into the other classes, I'd like to raise the question whether the consumer and producer classes are useful and really needed at all. Those classes set up systemd units starting cli tools contained in the Kafka package/tarball. The cli tools are are meant as examples, for testing and debugging imho though. Therefore, I doubt these classes are really useful at all. Of course just that we don't use it in our environment doesn't mean noone uses them, still I think this should be discussed.
Even the mirror class is questionable as Kafka has progressed a lot since the original mirrormaker implementation. But then, at least there's some usefulness in this so it makes more sense to keep them.

If we agree on that, I'll prepare the necessary changes to get rid of those classes.

Thanks for your opinions!

Best regards
Sven

TheMeier:
Since `kafka::mirror::config` relies on `kafka::consumer::config` and `kafka::producer::config` it would be easiest to get rid of them all at once.

PR #401, opened 2026-01-05
refactor: use system::mange_unit for broker service

TheMeier:
leverage systemd module instead of using erb templates. Will do some more follow ups for other units (if wee want to keep them https://github.com/voxpupuli/puppet-kafka/issues/381)

### DIFF
--- a/manifests/broker/service.pp
+++ b/manifests/broker/service.pp
@@ -35,27 +35,67 @@ class kafka::broker::service (
       'KAFKA_OPTS'       => $opts,
       'LOG_DIR'          => $log_dir,
     }
-    $environment = deep_merge($env_defaults, $env)
+    $environment = deep_merge($env_defaults, $env).map |$k, $v| { "'${k}=${v}'" }
 
     include systemd
 
-    if ($service_restart) {
-      $config_notify = Service[$service_name]
-    } else {
-      $config_notify = undef
+    $start_flag = $daemon_start ? {
+      true    => '-daemon ',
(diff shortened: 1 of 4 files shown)


manifests/broker/service.pp
```diff
@@ -35,27 +35,59 @@
...
+      },
+      'LimitCORE'        => $limit_core,
+      'LimitNOFILE'      => $limit_nofile,
+      'TimeoutStopSec'   => $timeout_stop,
+    }.filter |$k, $v| { $v != undef }
+
+    systemd::manage_unit { "${service_name}.service":
+      ensure          => 'present',
+      enable          => true,
+      active          => $service_ensure == 'running' ? {
```
TheMeier: hm strange I thought there was a linter that prohibts selectors in parameters

bastelfreak: yes thought so as well. or maybe that was during puppet-lint 4.x :thinking:

bastelfreak (approved):

manifests/broker/service.pp
TheMeier: Looks  like an issu in the linter.

this is ok: ` => $service_requires.empty ? {` this is ok `$service_ensure == 'running' ? {` but this is NOT ok: `  => $exec_stop ? {`

So I guess I will move all selectors out

TheMeier: maybe it this https://github.com/rodjek/puppet-lint/issues/840

PR #402, opened 2026-01-09
Feature/command config

bovy89:
If kafka has been configured with SASL and self-signed SSL certificates and using zookeeper is not an options (kraft is being used), you need to pass `--command-config` in order to be able to manage topics.

`/opt/kafka/bin/kafka-topics.sh --bootstrap-server kafka01.example.com:9093 --command-config /opt/kafka/config/admin.config --describe`

`/opt/kafka/config/admin.config`:

```
sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="<%= $kafka_admin_user %>" password="<%= $kafka_admin_password %>";
security.protocol=SASL_SSL
sasl.mechanism=PLAIN
<%- if $self_signed_certs { -%>
ssl.truststore.location=<%= $kafka_keystore_path %>
ssl.truststore.password=<%= $jks_password %>
ssl.truststore.type=JKS
<%- } -%>
```

### DIFF
--- a/manifests/topic.pp
+++ b/manifests/topic.pp
@@ -34,6 +34,11 @@
 #   A topic configuration override for the topic being created or altered.
 #   See the Kafka documentation for full details on the topic configs.
 #
+# @param cmd_config
+#   Property file containing configs to be passed to Admin Client.
+#   This is used only with --bootstrap-server option for describing and
+#   altering broker configs (e.g. specify sasl and ssl configs)
+#
 define kafka::topic (
   Optional[String[1]] $ensure                 = undef,
   Optional[String[1]] $zookeeper              = undef,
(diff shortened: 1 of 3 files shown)


manifests/topic.pp
```diff
@@ -52,10 +58,18 @@
...
+
   if $zookeeper {
     $_connection = $_zookeeper
   } else {
-    $_connection = $_bootstrap_server
+    if $cmd_config {
+      $_connection = "${_bootstrap_server} --command-config ${cmd_config}"
+    } else {
+      $_connection = $_bootstrap_server
+    }
```
bastelfreak: We can shorten this a bit:
```suggestion
    $_connection = if $cmd_config {
       "${_bootstrap_server} --command-config ${cmd_config}"
    } else {
      $_bootstrap_server
    }
```

TheMeier (approved):

PR #403, opened 2026-01-20
feat: use semicolon as delimiter when super.users config is passed as array

TheMeier:
also convert properties template to EPP

https://kafka.apache.org/41/security/authorization-and-acls/#_changing-the-default-behavior_

> note that the delimiter is semicolon since SSL user names may contain comma


bastelfreak (approved):
change looks good to me, but I'm not a kafka user at the moment

lbetz (approved):

